Issue #235, opened 2025-01-14
FEAT: Parse bits


joszamama:

andreas-zeller:
We need this for parsing binary files, such as GIF, which we again need for the Fandango paper.

andreas-zeller:
Completed in the `andreas` branch, together with a new `parse` command to test things. Try

```
$ echo -n 'a' | fandango parse -f docs/bits.fan --format=grammar -
<start> ::= <format_flag>
  <format_flag> ::= <italic> <bold> <underlined> <strikethrough> <brightness>
    <italic> ::= <bit>
      <bit> ::= 0
    <bold> ::= <bit>
      <bit> ::= 1
    <underlined> ::= <bit>
      <bit> ::= 1
    <strikethrough> ::= <bit>
      <bit> ::= 0
    <brightness> ::= <bit> <bit> <bit> <bit>
      <bit> ::= 0
      <bit> ::= 0
      <bit> ::= 0
      <bit> ::= 1
```

andreas-zeller:
This should now work well enough to be flagged as completed.

PR #284, opened 2025-01-26
Support for parsing with bit grammars

andreas-zeller:
* Fandango can now properly parse and decompose inputs using grammars with `<bit>` elements (#235). This is required for GIF support. Bits must come in multiples of eight, but that constraint should not be too hard.
* The Fandango CLI has a new `parse` command to test the parser.


joszamama:
Thank Andreas! I will take care of merging and testing it. (I have a meeting with two faculties from the University of Sevilla this morning, it will take a bit)

PR #311, opened 2025-02-06
Andreas

joszamama:


andreas-zeller:
Fixed

joszamama:
`make test` does work now, evaluation is still failing:

```
fandango:ERROR: Error in fix_individual: 
fandango:ERROR: Error in fix_individual: 
fandango:ERROR: Error in fix_individual: 
================================                                                                                                                                                                                                                                                                                            
XML Evaluation Results
================================
Total inputs: 200
Valid XML solutions: 104 (52.00%)
Grammar coverage (0 to 1): 0.64 (267 / 417)
Mean length: 18.62
Median length: 21.00
```

joszamama:
Problem is in function `fix_individual`:

```
suggested_tree = self.grammar.parse(
    value, failing_tree.tree.symbol
)
```

andreas-zeller:
Check for calls `.to_repr()` (or `repr()`) in derivation trees. These used to return a string, but now return an internal representation. Use `str()` instead.

joszamama:
Why do we need that change? There are 53 instances of `repr(` that have not been checked.

andreas-zeller:
Because it was wrong by Python conventions. Remember that all tests passed. We can debug this together.

PR #338, opened 2025-02-12
Bugfix: Bit-parser can't parse alternatives

alex9849:
**Describe the bug**
Fandano doesn't check for alternatives when parsing bytes.
If you try the example below, the first parsing attempt will work and the second will fail.
The reason for this is that alternatives aren't checked. Fandango only tries to parse the input using the first available option and ignores any alternatives that come after that.

**To Reproduce**
Use this grammar:
```
<start> ::= (<bit>{7} 1) | (<bit>{6} 1 <bit>)
<bit> ::= 0 | 1
```

Run:
```
with open("the_grammar.fan") as f:
    grammar, constraints = parse(f, use_stdlib=False)
assert grammar.parse(b'\x01', NonTerminal("<non_zero_byte>")) is not None
assert grammar.parse(b'\x02', NonTerminal("<non_zero_byte>")) is not None
assert grammar.parse(b'\x00', NonTerminal("<non_zero_byte>")) is None
```

**Expected behavior**
All assertions should pass.

This PR fixes the mentioned issue and introductes an additional test for verification.


alex9849:
Also another fix here. Constraints that work on sets can cause a fittness value that exceeds 1.0. In some cases this leads to rentries being accepted even if some constraints are failing. I also fixed that with my last commit in here.

Issue #80, opened 2024-11-28
BUG: Repetitions can currently be non-integral


addisoncrump:
https://github.com/fandango-fuzzer/fandango/blob/ef8c5e35443953f9de36a8dd429dadb0f1ffc8cd/language/FandangoParser.g4#L40-L41

This was caught by a type check in Rust: the repetition here specifies `NUMBER`, when it should specify `INTEGER`.

andreas-zeller:
Changing `NUMBER` to `INTEGER` in `FandangoParser.py` causes parsing errors; the repetition tests in `tests/test_grammar.py` fail due to a parse error. No idea why this should be the case. 

However, we sure can live with a slightly overly permissive grammar at this point, so I am assigning a low priority.

PR #347, opened 2025-02-26
Computed length repetitions

alex9849:
Adds dynamic computation of min and max value of repetitions. Works for generating and parsing.
Also added parsing of parameters to parser. A nonterminal might got one or more parameters in the grammar, that should be assigned to the derivation tree, when the Derivation tree got generated using that rule. The parser can now also set these parameters in the DerivationTree while parsing.


andreas-zeller:
Hi [alex9849](https://github.com/alex9849) – can you add test cases that document and show off how this works? 

I will need to update the tutorial accordingly, so I need to gain some understanding, especially related to the "parsing parameters." Generally speaking, I'd prefer properties to be determined from the resulting derivation tree rather than set during parsing, so please come up with a use case here too.

alex9849:
I added a testcase for parsing dynamic repetitions here: [For parsing](https://github.com/fandango-fuzzer/fandango/pull/347/files#diff-0c76c033dbce347029f7291ebb3db59b16c982432f9ccabb0b308a0c2e1383d8R205) [For generating](https://github.com/fandango-fuzzer/fandango/pull/347/files#diff-e4667073752b1036b6ea3454b46d6224c9137197d5e970e82da63d82867d04aaR78)

[This](https://github.com/fandango-fuzzer/fandango/pull/347/files#diff-842828d857611ab0b4eec3eae3eda2333788929d9162d150db0da282929692e8R3) is the grammar that both tests use. You may be wondering why I added a min(<nt>, 4) to both repetitions. The problem is that when Fandango fuzzes things, it generates the <len> tag first, because it comes first in the grammar. But the number that comes out there can be very high. Constraints aren't checked until after fuzzing, so the constraint that should prevent such high numbers hasn't been applied yet.


The parameter parsing part is not currently used in this version of fandango. I programmed this because I needed something like it on my fork. I've made some changes to my fork, so I don't need it anymore. But I might need it in the future. And since the parser code has a very strong code cohesion, I would like to keep it here for the future. At least for now. It doesn't affect Fandango's parsing capabilities.

A use case could be a parameter syntax like this:
```
<non_terminal> ::= <param_a:param_b:other_nt>
 # maybe In a future feature
<non_terminal> ::= <other_nt p_a='foo' p_b='bar'>
```
Here we have added some parameters to a rule. `<other_nt>` could theoretically be called from multiple rules with different parameters. 
In some fork or future feature, these parameters might be important for some behaviour that fandango performs. Also, the user might need them for something they do with the derivation tree returned by fandango. As a user, I added some parameters to my rule and I want to be able to access them after fuzzing so that I can continue to work with them. The same should apply to parsing. I have an output string and would like to see what parameters that output would have in its derivation tree form.

andreas-zeller:
As a side effect, this also fixes #80

andreas-zeller:
This is neat, but it currently fails

* `tests.test_grammar.ConstraintTest testMethod=test_repetitions_slice` (`make tests`), and
*  `docs/_build/html/ISO8601.md` (`make html`)

You may want to look into this before merging the pull request.

andreas-zeller:
I added text to the tutorial that discusses this

andreas-zeller:
> You may be wondering why I added a `min(<nt>, 4)` to both repetitions. The problem is that when Fandango fuzzes things, it generates the `<len>` tag first, because it comes first in the grammar. But the number that comes out there can be very high. Constraints aren't checked until after fuzzing, so the constraint that should prevent such high numbers hasn't been applied yet.

A downside of this might be that the `min()` function is also invoked during parsing, restricting the range of values that can be processed. I found that this is best handled by a generator, for instance:

```
<length> ::= <int> := str(random.randrange(1, 10))
```

Since generators are invoked only during production, this allows for full flexibility during parsing.

alex9849:
> This is neat, but it currently fails
> 
>     * `tests.test_grammar.ConstraintTest testMethod=test_repetitions_slice` (`make tests`), and
> 
>     * `docs/_build/html/ISO8601.md` (`make html`)
> 
> 
> You may want to look into this before merging the pull request.

Both issues should be fixed now. :)

Issue #181, opened 2024-12-17
Allow generators to have symbols as arguments


andreas-zeller:
Allow generators to have other grammar elements as arguments. This could be used to tie abstraction layers to each other, as in

```
<chunk> ::= <header> <compressed_content> <trailer>;
<compressed_content> ::= <byte>* := compress(<content>);
<content> ::= <foo> <bar> <baz> := uncompress(<compressed_content>);
```

where `<compressed_content>` could thus be derived from `<content>`, and vice versa.

andreas-zeller:
I understand this is now (partially?) implemented (#366); and I am currently writing a tutorial for this feature. If I take the following spec which encodes (and decodes?) data:

```python
import base64

<start> ::= b'Data: ' <item>
<item> ::= rb'.*' := base64.b64encode(bytes(<data>))
<data> ::= b'Fandango' <byte>+ := base64.b64decode(bytes(<item>))
```

Then, _fuzzing_ inputs works as expected:

```shell
$ fandango fuzz -f encode.fan -n 1  
Data: RmFuZGFuZ2/B                                                                                                                                                                                                            
```

However, _parsing_ inputs does not produce a `<data>` element as would be expected:

```shell
$ echo -n 'Data: RmFuZGFuZ2+euSVysw==' | fandango parse -f encode-decode.fan -o - --format=grammar
<start> ::= b'Data: ' <item>  # Position 0x0000 (0); b'Data: RmFuZGFuZ2+euSVysw=='
  <item> ::= b'RmFuZGFuZ2+euSVysw=='  # Position 0x0006 (6)
```

I am a bit stuck here, as the description in #366 is very scarce as it comes to parsing.

* How do we need to update the `parse()` function (or the `parse` command)?
* Where should the `<data>` elements be parsed into? `_children`? `_generator_params`? If not, where should one find them?
* The `DerivationTree` output (`to_tree()` etc.) does not include `_generator_params` elements. Should it?
* Can we access `<data>` elements in constraints?

and

* Can we update the `DerivationTree` code to document the interface, notably the   `generator_params` parameter?

@alex9849 – this is your call :-)

alex9849:
The CLI is actually something I overlooked when integrating the parameter feature into the codebase.
I have just written a fix for this here: https://github.com/fandango-fuzzer/fandango/tree/fix/show_generator_parameters_in_cli
This now:
- Updates the parse command to also derive generator parameters
- Updates the to_grammar() method to print generator parameters

[...]

PR #348, opened 2025-02-26
Support for grammar elements as arguments

w8floosh:
This PR implements the feature requested in issue #181 and adds a simple test for it.

Grammar elements can now be used inside generator positional or keyword arguments atomically, as operands or inside other expressions. 
The grammar elements are parsed as base selectors and the generator call includes the selector identifier, then the symbol is fuzzed and the identifier-value pair is added as a local variable for the eval() function, to allow injection of the generated string as a constant parameter.
These symbols are included in the USED_SYMBOLS set.
Usage in starred expressions is not yet supported as I could not think about real use cases for them.
The feature is provided with a test case.

Generator objects are now paired with RuleSearch objects when the generator contains grammar elements (see LiteralGenerator in language/grammar.py).

Visitors for the following parser rules were added: "arguments", "args", "arg", "kwargs".


joszamama:
Hi @w8floosh,

Thank you very much for your pull request!

We are going to integrate these changes into an intermediate branch in order to look for corner cases. I think your solution works, but I want to be sure it generalizes for different cases.

Issue #346, opened 2025-02-24
BUG: Some errors with expressions


andreas-zeller:
Consider this grammar (also in `docs/expr.fan`):

```
<start>   ::= <expr>
<expr>    ::= <term> " + " <expr> | <term> " - " <expr> | <term>
<term>    ::= <term> " * " <factor> | <term> " / " <factor> | <factor>
<factor>  ::= "+" <factor> | "-" <factor> | "(" <expr> ")" | <int>
<int>     ::= <digit>+
```

This invocation fails to find a solution:

```
$ fandango fuzz -f docs/expr.fan -n 10 -c 'len(str(<int>)) > 5'
```

Interestingly, this one works:

```
$ fandango fuzz -f docs/expr.fan -n 10 -c 'len(str(<int>)) == 5'
```

This invocation produces several solutions:

```
$ fandango fuzz -f docs/expr.fan -n 10 -c 'eval(str(<expr>)) == 10'
+05 / 96 + 10
+06 / 06 + 10
....
```

However, they are all invalid:

```
$ python
>>> +06 / 06 + 10
SyntaxError: leading zeros in decimal integer literals are not permitted; ...
```

These are demos I have given a number of times, so I think this is a regression.

andreas-zeller:
Once this is fixed, be sure to add test cases to avoid regressions.

TimScheckenbach:
The bug for the constraint `eval(str(<expr>)) == 10` is fixed with an additional test case. Regarding the constraint `len(str(<int>)) > 5`, as it is right now, there is no real solution for this, because in `grammar.py`, the maximal number of repetitions is bounded by `MAX_REPETITIONS`. I increased the bounds, but that only shifts the problem, so coming up with another solution might be better.

joszamama:
There is a new parameter for setting the max number of repetitions now added to the algorithm. I will take this into account for the new version of the algorithm I am currently working on.

PR #354, opened 2025-03-04
Fix and test for #346 

TimScheckenbach:


TimScheckenbach:
It now also has the `--max-repetition` option (#356) and tests for it

PR #366, opened 2025-03-11
Generators can have grammar elements

alex9849:
This PR allows generators to have nonterminals as function parameters as described in #181.

Consider the following grammar, which generates palindromes:
```
def reverse_input(input):
    return input[::-1]

<start> ::= <number> <rev_number>
<rev_number> ::= <number_tail>{0, 2} <number_start> :: reverse_input(<source_number>.to_string())
<source_number> ::= <number> :: reverse_input(<rev_number>.to_string())

<number> ::= <number_start> <number_tail>{0, 2}
<number_start> ::= '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9'
<number_tail> ::= '0' | <number_start>
<byte> ::= <bit>{8}
<bit> ::= 1 | 0

where <source_number>.<number> == <start>.<number>
```

The palindrome has two parts. The first part is a random number and the second part is the same number but reversed. The second part is represented in the grammar above by the non-terminal `<rev_number>`. `<rev_number>` is created using a generator that reverses its input. The input to the generator is `<source_number>`.
A constraint specifies that `<number>` (the first part.) and `<source_number>` must be equal. So this constraint passes through the generator.

When a parameterised generator is executed, fandango will first generate its parameters. It can happen that a parameter also uses generators. These generators will also be invoked. Nesting generators is possible. If necessary, their parameters will be generated. If a generator needs a parameter that is already part of the path (from the current tree node to the root), the generator will be ignored to break the otherwise infinite loop of generators calling each other.

If a generator `G_A` (which generates `P_A`) contains a parameter `P_B`, then this parameter `P_B` must be derivable using a generator `P_B` which takes `P_A` as an argument. This allows Fandango not only to generate children from parameters, but also to generate parameters from children.

Generators can have multiple nonterminal parameters. Imagine this:
```
<a> ::= ... := gen_A(<b>, <c>)
<b> ::= ... := gen_B(<a>, <c>)
<c> ::= ... := gen_C(<a>)
```
When generating `<a>`, Fandango uses the generator `gen_A(<b>, <c>)`. For this it first generates `<b>` and `<c>`. In case fandango needs to go the other direction and needs to derive the parameters for `gen_A(<b>, <c>)` from the already existing children, it will do the following. Fandango first looks at the generators for `<b>` and `<c>` and checks their attributes. From there it sees, that `gen_C` depends on `<a>`, and `gen_B` depends on `<a>` and `<c>`. In the beginning Fandango only knows `<a>` as this is the child for which we want to generate the parameters. So Fandango topologically sorts the required nonterminals according to their dependencies. This means Fandango will first run `gen_C(<a>)` to derive parameter `<c>` and then use the now known `<c>` to execute `gen_B(<a>, <c>)` to derive parameter `<b>`.

Good to know:
- DerivationTree now contains an attribute `generator_params` which stores used nonterminals if the children of a DerivationTree have been generated using a generator that uses attributes.
- Searches will also search in `generator_params`.
- DerivationTree.replace now requires the grammar as an attribute to keep `generator_params` in sync with `children` when running.
- DerivationTree.replace can also replace subtrees in `generator_params`
- grammar.parse() does not infer generator parameters. (Because of missing context and a recursive infinite loop that would occur.) If generator parameters are needed, first add the new subtree to the parent tree (if needed) to establish context, and call grammar.populate_generator_params(tree) for either the root, or the subtree afterwards. Note that manually adding a child (using set_children) doesn't update the generator parameters of DerivationTree. DerivationTree.replace does.


andreas-zeller:
This (mostly) addresses #181

PR #378, opened 2025-03-15
Fix/invalid tree size

alex9849:
- Fixes a bug where a tree size change wasn't propagated to the tree root.
- Fixes a bug where length tags used for computed length repetitions were not set to read-only.


joszamama:
Hi Alexander, I have just checked the branch and the tests are not passing. Can you take a look at them?

PR #409, opened 2025-04-07
Add soft constraints

leonbett:
This PR adds soft constraints in addition to the existing hard constraints (#4).

Soft constraints can be one of:
- `SoftValue`, i.e., a user-provided fitness score (not necessarily normalized), which is to be optimized. This could be something like "code coverage".
- `SoftConstraint`, i.e., a normal tree constraint that is to be optimized (i.e., try to satisfy the constraint in more contexts).

Fandango only tries to optimize soft constraints once it satisfied 100% of the hard constraints.

Fitness scores of soft constraints are normalized using t-digest based on prior observations.
Two keywords are introduced to the language: `max` and `min`.
For examples, see tests/test_softconstraint.py.


leonbett:
> Hi Leon – cool work! Could you please check (and possibly ensure with tests) that
> 
>     * there are no conflicts with `min` and `max` functions used in other contexts? (say, `where min min(<foo>, <bar>)` or `def f(x, y): return min(x, y) + max(x, y)`
> 
>     * one can specify such soft constraints on the command line, e.g., `fandango -f foo.fan -c 'min <age>` (or should we have a special option `--min <age>`?)

Hi Andreas,

Good point and thanks for your input! I've fixed the min/min issue and added a couple of tests.

To maximize `<age>`, i.e., a user-provided fitness score, you would have to use "fitness".
Without "fitness", Fandango uses its own fitness score, which can also be maximized, but in that case it tries to maximize the number of contexts in which a universally quantified constraint holds.

Some examples:

```
fandango fuzz -f docs/persons.fan -c 'max fitness int(<age>)'
fandango fuzz -f docs/persons.fan -c 'min fitness min (len(<name>), int(<age>)+20)' -N 100
```

andreas-zeller:
How about a `fitness()` _function_ that represents the built-in fitness score(s)? Then users may combine things as they like, as in 

```
maximizing fitness() + len(<input>) / 100
```

@leonbett: Let's discuss the alternatives tomorrow morning in a call; just ping me.

andreas-zeller:
Okay, so @leonbett and I settled on a "keyword" syntax:

```
<start> ::= ('a' | 'b' | 'c')+
where str(<start>) != 'd'
maximizing len(<start>)
```

The keyword (`where`, `maximizing`, `minimizing`) then becomes part of a constraint:

```
$ fandango -c "where str(<start>) != 'd'" -c "maximizing len(<start>)"
```

`where` is the default; it can be omitted in a constraint:

```
$ fandango -c "str(<start>) != 'd'" -c "maximizing len(<start>)"
```

Internally, constraints would continue to be handled as strings. When (un)parsing a constraint, prefixes `where`, `maximizing`, `minimizing` would have to be recognized and produced.

In order to have users discover this capability (and not having to remember keywords), the CLI should offer two options `--maximize C` and `--minimize C` that are equivalent to `-c "maximizing C"` and `-c "minimizing C"`, respectively.

There should be two explicit shorthands `--max` / `--min` (implicit ones will clash with algorithm settings):

```
$ fandango -c "str(<start>) != 'd'" --max "len(<start>)"
```

(The other shorthand alternative, `-m` (min) / `-M` (max), was dismissed as it is hard to disambiguate; and in the Python world, `-m` would typically refer to modules.)

And here we go!

leonbett:
Please don't merge yet, I am looking into some enhancements.

leonbett:
Looks like everything's done — let me know if anything else is needed.

joszamama:
Ok, I'll check the PR. Thanks Leon!

PR #443, opened 2025-05-06
Targeted mutations and fixes

### LINKED ISSUE #424: Mutation is not working as expected
### What it currently does

Consider the following grammar:

```
<start> ::= <digits>
<digits> ::= <digit> <digits> | <digit>
```

and the following input derived from it:
`010203040506070809`

Suppose this input is selected for mutation, and the failing tree (i.e., the subtree that Fandango will mutate) is `<digit>-<_digit>-0`.

Now [mutate() will replace](https://github.com/fandango-fuzzer/fandango/blob/c456c504f637e4e2a9d628592ab0d2e48ab9bef2/src/fandango/evolution/mutation.py#L67) every instance of the failing tree with a (the same) randomly generated tree.

Suppose the randomly generated tree is `<digit>-<_digit>-9`.

So the result of the mutation would be:
`919293949596979899`

### What I propose it should be doing

It should only replace one instance of the failing subtree. Ideally, at the precise location of the failing tree.

A possible result could be:
`010203940506070809`

leonbett:
- Enable targeted mutations. Previously, all instances of a given subtree were replaced by a mutation.
Targeted mutations are implemented as follows: We map derivation trees that are equal by value to the same reference using a dictionary. For instance, crossover or mutation may produce a new derivation tree (with a new reference) which was seen previously (with a different reference) and for which failing trees were already cached (by value, but mapping to a reference). Failing trees are required for targeted mutations. DerivationTree currently does not allow to uniquely identify nodes in a tree, e.g., by a path. Using unique references, we can now compare failing (sub-)trees with the current tree.
- Avoid 100% crossover rate because populations may have too little diversity to fill the population leading to a potential infinite loop.
- Fix a bug where `self.evaluation` was not updated to reflect the elite selection.
- Minor fixes

#424

### DIFF
--- a/src/fandango/evolution/adaptation.py
+++ b/src/fandango/evolution/adaptation.py
@@ -63,7 +63,7 @@ class AdaptiveTuner:
 
         # Adaptive Crossover
         if avg_diversity < diversity_low_threshold:
-            new_crossover_rate = min(1.0, self.crossover_rate * 1.05)
+            new_crossover_rate = min(0.9, self.crossover_rate * 1.05)
             LOGGER.info(
                 f"Generation {generation}: Increasing crossover rate from {self.crossover_rate:.2f} to {new_crossover_rate:.2f}"
             )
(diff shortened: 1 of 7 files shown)


leonbett:
Not ready for merge — I am looking into some issues

leonbett:
I re-implemented this in a different way in my Fandango fork. I'll upstream it later when I have some slack. Closing this for now.

PR #444, opened 2025-05-07
Add havoc_mutate: LibAFL's binary mutators

riesentoaster:
Usage:

```
<start> ::= <byte>* := havoc_mutate(<your_tree_start>)
```

This applies varying numbers of bit and byte level mutations on the serialized version of your tree, such as bit flips, replacing or extending your input with random values or other slices from the input, deleting parts of the input, etc. The mutations are a Python reimplementation of [LibAFL's `havoc_mutations`](https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/mutators/mutations.rs), the scheduling is based on [LibAFL's `HavocScheduledMutator`](https://github.com/AFLplusplus/LibAFL/blob/f33376f1cdb42955dccc2c73d160104167574b0a/libafl/src/mutators/scheduled.rs#L183).

Refer to the experiment `havoc` for an example of it being used on libgif, and an experiment setup to compare it against "pure" Fandango and `libafl_libfuzzer`.


src/fandango/language/tree.py
```diff
@@ -296,7 +296,14 @@ def to_string(self) -> str:
...
             # This is a bit value; convert to bytes
-            val = int(val).to_bytes(val // 256 + 1)
+            assert (
+                val >= 0
+            ), "Assumption: ints are unsigned. If this does not hold, the following needs to change"
+            required_bytes = (val.bit_length() + 7) // 8  # for unsigned ints
+            required_bytes = max(
+                1, required_bytes
+            )  # ensure at least 1 byte for number 0
+            val = int(val).to_bytes(required_bytes)
```
riesentoaster, line 306: This is a bug I've stumbled across during testing. I'm not entirely sure about the assumption about the int being unsigned. Can anyone confirm or deny this?

PR #454, opened 2025-05-20
Add docs, remove duplicate code, improve separation of responsibilities, fix types

riesentoaster:
I'm trying to understand the structure of the main algorithm, and I started refactoring certain things as I understood them better, removed (almost) identical code, extracted functions, moved code around so it better fits, etc.

Also: The following types are deprecated (see [Python docs](https://docs.python.org/3/library/typing.html#typing.Dict)) in favour of their lowercase counterparts, so I unified their usage across the code base:
- `List` -> `list`
- `Dict` -> `dict`
- `Set` -> `set`
- `Tuple` -> `tuple`
- etc.

### DIFF
--- a/src/fandango/evolution/population.py
+++ b/src/fandango/evolution/population.py
@@ -1,4 +1,4 @@
-from typing import Callable, List, Optional, Set, Tuple
+from typing import Callable, Optional, Set, Tuple
 
 from fandango.constraints.fitness import Comparison, ComparisonSide, FailingTree
 from fandango.language.grammar import DerivationTree, Grammar
@@ -22,7 +22,7 @@ class PopulationManager:
 
     @staticmethod
     def add_unique_individual(
-        population: List[DerivationTree],
+        population: list[DerivationTree],
         candidate: DerivationTree,
         unique_set: Set[int],
     ) -> bool:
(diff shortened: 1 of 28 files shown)


src/fandango/evolution/population.py
```diff
@@ -20,30 +20,57 @@ def __init__(
...
+        :param candidate: The individual to potentially add to the population.
+        :param unique_set: The set of unique individuals.
+        :return: True if the individual was added, False otherwise.
+        """
         h = hash(candidate)
         if h not in unique_set:
             unique_set.add(h)
             population.append(candidate)
             return True
         return False
```
riesentoaster, line 42: I used this in a couple of more places that did the same thing.

Is there a reason we're not just using a simple set of `DerivationTree`s that is converted to a list when needed instead of a parallel list and set of hashes? Is it faster?

src/fandango/evolution/population.py
```diff
@@ -20,30 +20,57 @@ def __init__(
...
```
riesentoaster: I used this in a couple of more places that did the same thing.

Is there a reason we're not just using a simple set of `DerivationTree`s that is converted to a list when needed instead of a parallel list and set of hashes? Is it significantly faster?

joszamama: If I remember correctly, if we deal with huge derivation trees, this process gets too expensive. With hash reprs we get a better performance.

PR #459, opened 2025-06-03
Add infinite generation mode

riesentoaster:


riesentoaster:
TODOs:
- [ ] Make `_evolve_io` call `generate` instead of `_evolve_single`
- [ ] Output solutions as they come up in infinite mode

PR #460, opened 2025-06-06
IO Grammar

alex9849:
Add support for IO Grammars to


src/fandango/constraints/predicates.py
```diff
@@ -1,4 +1,5 @@
-from fandango.language.tree import DerivationTree
+from fandango.language.symbol import NonTerminal
+from fandango.language.io import *
```
riesentoaster: Can we use explicit imports or at least something like `import fandango.language.io as io`?

src/fandango/constraints/predicates.py
```diff
@@ -65,3 +66,20 @@ def is_after(
     Check if the tree is after the after_tree and before the before_tree.
     """
     return is_before(tree, before_tree, after_tree)
+
+
+def get_index_within(
+    tree: DerivationTree, scope: DerivationTree, index_counter_symbols: list[str]
+):
```
riesentoaster, line 73: Add type hints for return type. Consider returning `None` if nothing was found instead of -1 to be more explicit. The return type would then be `Optional[int]`.

src/fandango/constraints/predicates.py
```diff
@@ -65,3 +66,20 @@ def is_after(
...
+    index_counter_nts = [NonTerminal(symbol) for symbol in index_counter_symbols]
+    for val in scope.flatten():
+        if val == tree:
+            return idx
+        if val.symbol in index_counter_nts:
+            idx += 1
+    return -1
+
+
+def get_type(tree: DerivationTree):
```
riesentoaster, line 84: More return type annotations missing

src/fandango/evolution/algorithm.py
```diff
@@ -306,7 +320,160 @@ def _perform_destruction(
...
+    def fix_individual(
+        self, individual: DerivationTree
+    ) -> tuple[float, DerivationTree]:
+        _, failing_trees = self.evaluator.evaluate_individual(individual)
+        replacements = dict()
+        for failing_tree in failing_trees:
+            if failing_tree.tree.read_only:
+                continue
+            for operator, value, side in failing_tree.suggestions:
+                from fandango.constraints.fitness import Comparison, ComparisonSide
```
riesentoaster: Why do we have imports in the middle here? Put them at the top?

src/fandango/evolution/algorithm.py
```diff
@@ -306,7 +320,160 @@ def _perform_destruction(
...
+                if operator == Comparison.EQUAL and isinstance(
+                    value, (str, bytes, DerivationTree)
+                ):
+                    if side == ComparisonSide.RIGHT:
+                        continue
+                    if (
+                        isinstance(value, DerivationTree)
+                        and failing_tree.tree.symbol == value.symbol
+                    ):
+                        suggested_tree = value.__deepcopy__(None, True, False, False)
```
riesentoaster: I think you're not supposed to use dunder methods manually

src/fandango/language/tree.py
```diff
@@ -527,35 +649,36 @@ def split_end(self) -> "DerivationTree":
...
+    def replace_multiple(
+        self, grammar: "Grammar", replacements: dict["DerivationTree", "DerivationTree"]
+    ):
         """
         Replace the subtree rooted at the given node with the new subtree.
         """
-        if self == tree_to_replace and not self.read_only:
-            new_subtree = deepcopy(new_subtree)
+        if self in replacements and not self.read_only:
+            new_subtree = replacements[self].__deepcopy__(None, True, False, False)
```
riesentoaster, line 681: More manual calls to dunder methods

src/fandango/language/tree.py
```diff
@@ -527,35 +649,36 @@ def split_end(self) -> "DerivationTree":
...
-
     def replace(self, grammar: "Grammar", tree_to_replace, new_subtree):
+        return self.replace_multiple(grammar, {tree_to_replace: new_subtree})
+
+    def replace_multiple(
+        self, grammar: "Grammar", replacements: dict["DerivationTree", "DerivationTree"]
+    ):
         """
         Replace the subtree rooted at the given node with the new subtree.
         """
```
riesentoaster, line 679: This is no longer correct, right?

src/fandango/evolution/algorithm.py
```diff
@@ -306,7 +320,160 @@ def _perform_destruction(
...
+        else:
+            receiver = "*" + receiver
+
+        if msg.contains_bytes():
+            msg = msg.to_bytes()
+        else:
+            msg = msg.to_string()
+
+        LOGGER.info(f"({sender} -> {receiver}): {msg}")
+
```
riesentoaster: Move this to the logger class?

src/fandango/evolution/algorithm.py
```diff
@@ -306,7 +320,160 @@ def _perform_destruction(
...
     def evolve(self) -> list[DerivationTree]:
+        if self.grammar.fuzzing_mode == FuzzingMode.COMPLETE:
+            return self._evolve_single()
+        elif self.grammar.fuzzing_mode == FuzzingMode.IO:
+            return self._evolve_io()
+        else:
+            raise RuntimeError(f"Invalid mode: {self.grammar.fuzzing_mode}")
+
+    def _evolve_io(self) -> list[DerivationTree]:
```
riesentoaster, line 336: Add a couple of inline comments to explain what is happening in this method. If a new user wants to understand the algorithm and architecture of FandangoIO, they will read this function and try to understand it. We can make that significantly easier with a couple of comments.

Plus, obviously, good function documentation.

src/fandango/evolution/algorithm.py
```diff
@@ -306,7 +320,160 @@ def _perform_destruction(
...
+        global_env, local_env = self.grammar.get_python_env()
+        io_instance: FandangoIO = global_env["FandangoIO"].instance()
+        history_tree: DerivationTree = random.choice(self.population)
+
+        self.desired_solutions = 1
+        while True:
+            self.population.clear()
+            self.evaluator.reset()
+            forecaster = PacketForecaster(self.grammar)
```
riesentoaster: Is there a reason we're instantiating this every loop?

riesentoaster (changes requested):
- [x] Check that all functions have type hints including return types
- [x] Check that all (public) functions are documented
- [x] Improve separation of concern: For non-trivial classes (i.e. classes that aren't just a fancy tuple), don't access variables directly. Write getters/setters or think about if the logic you need the data for should belong in the target class.
- [x] The same goes for multiple references of data. Only store each variable in one class. If you need to access the data in another class, write getters/setters or pass them as args to methods.
- [ ] Split excessively long files into submodules (such as `src/fandango/language/grammar.py`
- [x] Add at least rudimentary test cases
- [x] Fix failures from existing tests

src/fandango/language/io.py
```diff
@@ -0,0 +1,393 @@
+import enum
+import select
+import socket
+import sys
+import threading
+import time
+from typing import Callable, Tuple
```
riesentoaster: Use the lowercase `tuple`

src/fandango/language/io.py
```diff
@@ -0,0 +1,393 @@
...
+    SERVER = ("Server",)
+    CLIENT = "Client"
+
+
+class IpType(enum.Enum):
+    IPV4 = ("IPv4",)
+    IPV6 = "IPv6"
+
+
+class Ownership(enum.Enum):
```
riesentoaster, line 27: Why not just `Owner`?

src/fandango/evolution/algorithm.py
alex9849: In this case I do because there are some parameters in there that I added to allow optimizations.

src/fandango/language/io.py
```diff
@@ -0,0 +1,393 @@
...
+import sys
+import threading
+import time
+from typing import Callable, Tuple
+
+from fandango import FandangoError
+from fandango.language.tree import DerivationTree
+
+
+class ProtocolType(enum.Enum):
```
riesentoaster, line 12: No need to have `Type` in type names, we already know they are types

src/fandango/evolution/algorithm.py
```diff
@@ -306,7 +320,160 @@ def _perform_destruction(
...
+                        new_packet.recipient
+                    ].is_fuzzer_controlled()
+                ):
+                    io_instance.set_transmit(
+                        new_packet.sender, new_packet.recipient, new_packet.msg
+                    )
+                    self.log_message_transfer(
+                        new_packet.sender, new_packet.recipient, new_packet.msg, True
+                    )
+                    exec("FandangoIO.instance().run_com_loop()", global_env, local_env)
```
riesentoaster: Can we send the message in the `set_transmit` function? And thus remove the exec? That would be much cleaner.

src/fandango/evolution/algorithm.py
riesentoaster: Rename the function to `deepcopy` or something like that, and call it with fixed values from `__deepcopy__`

src/fandango/evolution/algorithm.py
```diff
@@ -306,7 +320,160 @@ def _perform_destruction(
...
+
+            if len(forecast.getMsgParties()) == 0:
+                if len(history_tree.find_protocol_msgs()) == 0:
+                    raise RuntimeError("Couldn't forecast next packet!")
+                return [history_tree]
+
+            msg_parties = []
+            for party in forecast.getMsgParties():
+                if io_instance.parties[party].is_fuzzer_controlled():
+                    msg_parties.append(party)
```
riesentoaster: How about something like

```python
msg_parties = [party for party in forecast.getMsgParties() if io_instance.parties[party].is_fuzzer_controlled()]
```

I personally like not mutating variables if it doesn't simplify something considerably. It makes the code easier to understand. It's not something that is black-and-white wrong, and list comprehensions can become quite hard to understand as well if they get too complicated, but here I think it would be more readable.

The same applies to a couple other places as well.

src/fandango/evolution/algorithm.py
```diff
@@ -306,7 +320,160 @@ def _perform_destruction(
...
+
+                evolve_result = self._evolve_single()
+                if len(evolve_result) == 0:
+                    nonterminals_str = " | ".join(
+                        map(lambda x: str(x.node.symbol), fuzzable_packets)
+                    )
+                    raise RuntimeError(
+                        f"Couldn't find solution for any packet: {nonterminals_str}"
+                    )
+                next_tree = evolve_result[0]
```
riesentoaster, line 384: What if there is no message returned here?

src/fandango/evolution/algorithm.py
```diff
@@ -306,7 +320,160 @@ def _perform_destruction(
...
+                    forecast, io_instance
+                )
+                self.log_message_transfer(
+                    packet_tree.sender,
+                    packet_tree.recipient,
+                    packet_tree,
+                    False,
+                )
+
+                hookin_option = next(iter(forecast.paths))
```
riesentoaster, line 415: Does this always *have* to exist?

src/fandango/evolution/algorithm.py
```diff
@@ -430,6 +597,145 @@ def evolve(self) -> list[DerivationTree]:
 
         return self.solution
 
+    def msg_parties(self) -> list[FandangoParty]:
+        global_env, local_env = self.grammar.get_python_env()
+        io_instance: FandangoIO = global_env["FandangoIO"].instance()
+        return list(io_instance.parties.values())
+
+    def _parse_next_remote_packet(
```
riesentoaster, line 552: Move this to another file?

Also: Split this into multiple functions, it's way too long.

src/fandango/evolution/population.py
```diff
@@ -89,27 +101,26 @@ def refill_population(
...
                 failing_trees,
             )
-            if PopulationManager.add_unique_individual(
+            if not PopulationManager.add_unique_individual(
                 current_population, candidate, unique_hashes
             ):
                 attempts += 1
 
-        if len(current_population) < self.population_size:
+        if not self._is_population_complete(current_population):
```
riesentoaster, line 118: Nice catch!

src/fandango/evolution/algorithm.py
alex9849: Yes. The entry only exists, if it is not empty.

src/fandango/evolution/algorithm.py
alex9849: That case is taken care of in the code directly before that.

joszamama:
Also, please, take the time to document the main functions of FandangoIO

Issue #179, opened 2024-12-17
FEAT: A much faster parser


andreas-zeller:
Make use of the Python _built-in parser_ instead of using our own:

Split the `.fan` file into three parts:
1. Grammar rules, identified by `::=`
2. Constraints, identified by `;` or a new leading keyword (#180)
3. Python code

Then, use the Python parser to parse (2) and (3); use a lightweight parser to parse (1) (or convert (1) into something that can be parsed by Python).

Note:
* This would also reduce the need for caching parse results.
* This may require to implement our own `import` (#178)
* This could allow us to implement more grammar syntax features (#65)

smythi93:
I will link this issue to #374, #369, #177, and #175. To work on it in a single branch.

andreas-zeller:
#369 is already fixed (at least in my branch); #175 and #177 should not wait until we have a new parser

andreas-zeller:
Before we go and implement an all-new (recursive descent?) parser for this (and run into the problem of having to maintain two parsers), also consider the option of taking the existing ANTLR spec (which works well apart from minor bugs (#374), and which is complete) and translate it into a _C_ parser rather than a Python parser. All we'd need then is a converter that translates the C AST back into a Python AST, and from there, we can use the existing (and working) infrastructure.

andreas-zeller:
With the help of the [_speedy-antlr_ project](https://speedy-antlr-tool.readthedocs.io/), I managed to integrate a new `.fan` parser that uses ANTLR-generated C++ code. There are still bugs as it comes to processing Python code (we need to implement our own lexer in C++), but I can already tell this new parser is 17-20x faster.

PR #464, opened 2025-06-08
A much faster parser

andreas-zeller:
This pull request brings a new parser for `.fan` files implemented in C++ that is ~17x-20x faster than the (slow) Python parser. It is created from the same ANTLR sources as the Python parser and should be 100% compatible. This addresses #179.

Including C++ code into Fandango has a few consequences:

* Users may need to compile C++ code to set up Fandango
* We may need to build and distribute binaries for various platforms
* We may need to test these as part of our CI.

The good news is that if any of the above fails, the Python parser can and will be used as a fallback option. There is a `fandango`-level option `--parser` that helps select the parsing code to be used; `--parser=speedy` forces using the C++ code; `--parser=python` forces using the Python code. Default is `auto`, choosing the C++ code if available and falling back to Python otherwise. Setting the global `fandango.Fandango.parser` string variable to one of the above values has the same effect.

Technically, this is realized by having ANTLR generate C++ files; the [speedy-antlr-tool](https://speedy-antlr-tool.readthedocs.io/) links them to Python code. I had to implement a Python-specific lexer in C++. The whole process is transparent; we can still maintain all of Fandango syntax from the ANTLR files. Since caching continues to improve speed over parsing, we keep on caching parsed `.fan` files, even though the effect on speed is smaller now.

[Here is an example](https://github.com/amykyta3/speedy-antlr-example/blob/master/.github/workflows/build.yml) on how to integrate this into our build and CI systems; replace `SPAM_EXAMPLE_REQUIRE_CI_BINARY_BUILD` with `FANDANGO_REQUIRE_CI_BINARY_BUILD`.

As a side effect, this pull requests also brings

* extensive `.fan` parser testing (notably differential testing of the `.fan` parsers)
* a new `dump` CLI command to output a parsed `.fan` file again
* fixes for a number of minor bugs discovered during testing, mostly related to unparsing specs


joszamama:
Hi @andreas-zeller, two questions:

1. Should we move the --parser option into the fuzz and parse commands? Instead of `fandango --parser speedy fuzz -f demo/01-miller.fan`, we would do `fandango fuzz -f demo/01-miller.fan --parser speedy`
2. Should we rename the `--parser=speedy` option to `--parser=cplusplus`?

Let me know your thoughts on this :)

joszamama:
> The package that enables this is called `speedy-antlr-tool`, so `speedy` was a reasonable name. If you want a C++ reference, we could go for `cpp` instead.

Done, it is now `--parser=cpp`

Issue #439, opened 2025-05-02
ANTLR -> Fandango grammar converter


b-fein:
The AST 2025 keynote presentation by @andreas-zeller mentioned the possibility of converting ANTLR grammars into the fandango format[^1]. In https://github.com/fandango-fuzzer/fandango/tree/main/utils I found the converters for 010 and DTDs, but not for ANTLR.

I opened the issue since I did not find a related one yet. I would be interested in such a converter since it would make it easy to try fandango out to generate testing inputs for a custom parser implementation for a DSL in one of our tools.

[^1]: https://conf.researchr.org/details/ast-2025/ast-2025-papers/17/Personalized-Fuzzing, slide 44

andreas-zeller:
Yep, that's a project we have offered to a student team; if they jump on it, we should have something by July 2025.

If you can't wait, you can build a simple ANTLR to Fandango converter in 2-3 days:
  
1. Start with building an ANTLR parser for ANTLR (using [this definition](https://github.com/antlr/grammars-v4/tree/master/antlr/antlr4))
2. Use this to parse ANTLR files into an abstract syntax tree (AST)
3. Walk through the AST to turn ANTLR fragments and tokens into regexps, and syntactic rules into Fandango grammars.

The challenge lies in the details – converting ANTLR regexps into Python regexps, or dealing with states; but these can be addressed on the fly. So if someone reads this, feel free to contribute your own converter!

andreas-zeller:
Hi @b-fein – Today, I implemented a basic ANTLR → Fandango converter (and an associated `fandango convert` command) today; see PR #482.
This is now in testing and review; it will be part of the 1.0 release, which is due in two weeks from now.

PR #482, opened 2025-06-14
New `convert` command converts ANTLR specs and more to Fandango

andreas-zeller:
This pull request brings a new command, `fandango convert`  that allows converting input specifications from various other formats. For now, these are

* ANTLR
* 010 binary templates
* DTDs

For instance,

```shell
$ fandango convert Calculator.g4
```

takes an ANTLR spec `Calculator.g4` and produces an (almost) equivalent Fandango spec.

Note that this is still experimental. Each format has a few features that are hard to replicate in Fandango, and the converters have undergone only very basic testing. Still, the produced `.fan` files should serve as a good base for further refinement; and extensions to the converters are welcome!

This addresses #439; documentation to follow.


b-fein:
Thanks for working on this!

After manually editing the Fandango grammar to fix the issues below, the conversion from Antlr seems to work according an initial cursory review of the produced outputs (except for the noted not yet possible inversions of rules):

* The parser crashes for non-ascii characters in the input grammar (e.g., in my case `…` in a comment).

* Probably not specific to this pull request, but a general problem with the Fandango grammar parser. The fandango grammer parser does not seem to like rules that have the same name as internal keywords (?).
  For example, in our case we have a rule `not` in the Antlr grammar which gets converted to a Fandango grammar:
  ```
  <boolExpr> ::= <binaryBoolExpr> | <not> | [other rules omitted]
  <not> ::= 'not ' <exprOrLiteral>
  ```
  Fandango then exits with
  ```
  FandangoSyntaxError: no viable alternative at input '<boolExpr> ::= <binaryBoolExpr> | <not'
  ```
  Renaming it to `<notExpr>` fixes this issue, so I guess the use of the keyword `not` is the issue.

* Antlr grammars tend to use the built-in lexer rule `EOF` as marker for the end of input to force complete parsing of the input.
  This rule is missing from the Fandango grammar, resulting in a
  ```
  FandangoValueError: Undefined symbol <EOF> in grammar
  ```

Issue #374, opened 2025-03-15
No f-strings in Python code


andreas-zeller:
This `.fan` spec raises a `NameError: name 'x' is not defined`

```python
def fun():
    x = "y"
    print(f"{x}")

<start> ::= "a"
```

This is raised from a call to `eval()` in 
```
File "/Users/zeller/Projects/Fandango!/src/fandango/language/symbol.py", line 101, in clean
```

andreas-zeller:
See also `visitStrings()` in `src/language/convert.py`.

Issue #177, opened 2024-12-17
Have `**<foo>` produce a list of all `<foo>` elements in the population


andreas-zeller:
This allows quantification, as in

```
any(n.startswith("A") for n in **<name>)
```

and most notably, checking the fitness of the population:

```
has_gauss_distribution(**<start>)
```

leonbett:
Some thoughts on this:
- This requires extending the fitness calculation. Currently, Fandango assigns a fitness score to each individual. With the proposed feature, Fandango would additionally produce a score for the entire population.
- We generally have goals pertaining to individuals (validity, control-flow distance to a specific line of code) and the population (has_gauss_distribution, maximizing code coverage).
- When optimizing a population goal, the population size shouldn't be limited as it currently is; and some more changes to the algorithm would be required (e.g., we probably shouldn't throw away 90% of the population per generation).

Issue #175, opened 2024-12-17
Have `*<foo>` produce a list of all `<foo>` elements in an individual


andreas-zeller:
This allows quantification, as in 

```
any(n.startswith("A") for n in *<name>)
```

as well as checking for existence, as in

```
not *<name>
```

With this, we can eliminate `forall` and `exists` keywords (#110, #117)

PR #486, opened 2025-06-16
New Language Features

smythi93:
This PR addresses issues #177, #175, and #374.
Additionally, some smaller fixes and improved tests, e.g., relative path to the resources and closing file streams.

### DIFF
--- a/src/fandango/language/grammar.py
+++ b/src/fandango/language/grammar.py
@@ -2301,7 +2301,7 @@ class Grammar(NodeVisitor):
         """
 
         initial = set()
-        initial_work: [Node] = [NonTerminalNode(name) for name in self.rules.keys()]  # type: ignore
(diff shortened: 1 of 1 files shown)


.github/workflows/code-checks.yml
```diff
@@ -56,8 +56,8 @@ jobs:
           files: |
             src/**/*.py
             src/**.py
-            tests/**/*.py
-            tests/**.py
+          # tests/**/*.py no type checking for tests
+          # tests/**.py no type checking for tests
```
riesentoaster, line 60: Is there a reason we're not doing type checking for tests? They are using the code's interfaces, and if there is an issue where you can't really build tests without breaking type hints, chances are the types are wrong.

smythi93: Theoretically, yes, but practically, writing tests that conform to static typing in Python is sometimes challenging / most of the time unnecessary. Often, you purposely break mypy checks in tests to have better and more understandable tests. But if you want, currently we have 183 checks failing in 16 of our 20 test files...

PR #505, opened 2025-06-20
Introduce beartype

riesentoaster:
[Beartype](https://beartype.readthedocs.io) performs dynamic type checking. I set it up such that it is activated in the unit tests only. Just running it shows that our type hints aren't always correct, so I disabled it globally for now (see `src/fandango/__init__.py`). The idea here is to iteratively reduce the blocklist until we're down to just the autogenerated code (if necessary).

If performance becomes too much of an issue, we can maybe further reduce the runs it is active on.


joszamama:
Tests are not passing

PR #513, opened 2025-06-23
Unify dependency lists, build book in CI using Makefile script

riesentoaster:
This unifies where dependencies are stored to `pyproject.toml`, removes unnecessary dependencies and removes duplicated dependencies.


riesentoaster:
Depends on #515

PR #524, opened 2025-06-24
Fix Fandango IO CLI

riesentoaster:


src/fandango/evolution/algorithm.py
```diff
@@ -528,6 +436,114 @@ def generate(
...
+
+        while True:
+            forecast = forecaster.predict(history_tree)
+
+            if len(forecast.get_msg_parties()) == 0:
+                if len(history_tree.protocol_msgs()) == 0:
+                    raise FandangoFailedError("Could not forecast next packet")
+                yield history_tree
+                continue
+                # TODO: Reset for next iteration
```
riesentoaster, line 461: @alex9849 Could you implement this?

PR #533, opened 2025-06-25
Fix/530 500 max node not strict and recursion error

alex9849:
Fixes issues:
- 500 (RecursionError when sending expressions)
- 530 (max_nodes is not strict)


alex9849:
Did the CI just skip the tests?

PR #574, opened 2025-06-27
fixed constraint parsing

smythi93:


joszamama:
Hey @smythi93, there's a big number of tests failing, could you take a look at the reason?

PR #582, opened 2025-07-02
Computed length repetitions refactor

alex9849:
This PR introduces changes that refactor the current implementation of computed length repetitions, incorporating Fandango's approach of strictly separating syntactic generation from semantic constraint solving.


alex9849:
Still a draft. Don't merge yet.

PR #594, opened 2025-07-08
Refactor DerivationTree to Achieve Better Isolation of Symbol Values

riesentoaster:
In preparation of further out-of-grammar fuzzing advancements, I need better isolation of the values in `DerivationTree`s. Previously, all sorts of classes directly accessed it, thus making shadowed out-of-grammar mutation infeasible to implement reliably.

This PR primarily introduces `TreeValue`, a wrapper around the values of each symbol, including those of `DerivationTree`s. It simplifies combining values and has a well-defined interface to access its content, to ensure consistency between the different types it may contain (bits, bytes, strings).

I further reduced the direct access on the data from the outside interface on `DerivationTree`, to further reduce the likelihood of unexpected behavior because the underlying data structure changed. All functionality is still available (in fact, more control over encodings is now possible), but I'm forcing users to think about the datatype they expect by manually converting them to `str`s, `int`s or `bytes`. If there is a good reason to add additional ways of interfacing with `DerivationTree`s *with obvious behavior*, I'm open to discussion.

This is still a work in progress, cleanup is still not done, and tests are still failing.


joszamama (commented):
Reformat of types, no more implicit declarations, adapted some experiments to the new style, new `format_as_spec` method, typing

joszamama:
Can we move `conftest.py` to the `/tests` folder?

PR #602, opened 2025-07-17
Introduce grammar settings

riesentoaster:


andreas-zeller:
See my suggestion #605 on using Python to change Fandango settings in `.fan` files

PR #618, opened 2025-07-24
Re-introduce direct access functions for underlying DerivationTree/TreeValue items

### LINKED ISSUE #609: String methods no longer available on nonterminal symbols
In `dev`, constraints like `<name>.startswith("C")` produce the error

```
AttributeError: 'DerivationTree' object has no attribute 'startswith'
  Evaluation failed: ___fandango_4376668880_0___.startswith('C')
  Convert <symbol> to the expected type, say 'str(<symbol>)', 'int(<symbol>)', or 'bytes(<symbol>)'
```

This is a regression with respect to 1.0. A conversion should not be necessary; nonterminal symbols should inherit all string methods.

This is a standard example in the tutorial, but (apparently) not part of the test suite. (Maybe make the tutorial part of our tests? Or introduce automated regression tests?)

riesentoaster:
Such accesses will also present `DeprecationWarning`s with concrete instructions on what to change to make it compatible with the improved interface.

Closes #609 

The book/docs are now also *built* in PRs, but only *deployed* in pushes to `main`, so on merge. @andreas-zeller for some reason the `make html` command executed in the CI does not fail if any of the doc assertions fail (see run in this PR). Do you know why/what would need to be changed to check this? Also: There currently is an issue with the iterative parser, see #619.

### DIFF
--- a/evaluation/vs_isla/tar_evaluation/tar.fan
+++ b/evaluation/vs_isla/tar_evaluation/tar.fan
@@ -62,18 +62,28 @@ def generate_file_name():
         '5', '-', '*', 'Z', '?', '3', 'b', 'v', 'U', 'D', '"', '9',
         'M', 'i', 'm', '8', ']', 'x', '7', '.', 'B'
     ]
+    # Previous versions of Fandango supported creating DerivationTrees directly in generators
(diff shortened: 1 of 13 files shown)


riesentoaster:
Closed in favor of #630

PR #630, opened 2025-07-29
Re-introduce direct access functions for underlying DerivationTree/TreeValue items

### LINKED ISSUE #609: String methods no longer available on nonterminal symbols

riesentoaster:
Closes #609


docs/cat-oracle.fan
```diff
@@ -1,5 +1,5 @@
 <start> ::= <In:input> <Out:output>
 <input> ::= <line>
 <output> ::= <line>
-<line> ::= r'[.*\n]+'
+<line> ::= r'.*\n'
```
andreas-zeller, line 5: Check this. `cat` should be able to handle multiple lines (but then, the nonterminal should be called `<lines>`).

src/fandango/language/tree.py
```diff
@@ -77,6 +80,36 @@ def index_by_reference(lst: Iterable[T], target: T) -> Optional[int]:
     return None
 
 
+def forward_to_tree_value_methods(
+    function_names: list[str],
+):
+    """
+    Decorator to add dunder methods to a class, delegating to the result of `value().method(self)`.
```
andreas-zeller: `dunder`?

PR #633, opened 2025-07-31
Split constraints module into more files

riesentoaster:
This does one thing, and one thing only: Split the two files from the `constraints` module into shorter files that have better inner cohesion, while ensuring that imports remain correct and tests still pass.

*No logic was changed*


pyproject.toml
```diff
@@ -114,4 +114,4 @@ exclude = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "-n auto --timeout 700"
+addopts = "-n auto --timeout 700 --durations=10"
```
riesentoaster, line 117: This will print the slowest 10 tests, which means that if someone gets annoyed by how slow the tests are, they know where to start optimizing. :D

PR #650, opened 2025-08-12
feat: add configurable max caps to AdaptiveTuner to prevent unbounded…

joszamama:
… parameter growth and crashes

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -3,6 +3,8 @@ name: Build & Upload to PyPI and Docker Hub
 on:
   release:
     types: [published]
+  pull_request:
+    branches: [release]
   workflow_dispatch:
 
 jobs:
@@ -25,7 +27,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.11.2
+        uses: pypa/cibuildwheel@v3.1.3
         env:
           FANDANGO_REQUIRE_BINARY_BUILD: "1"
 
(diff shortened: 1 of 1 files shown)


src/fandango/evolution/adaptation.py
```diff
@@ -36,71 +41,90 @@ def update_parameters(
...
-                f"Generation {generation}: Decreasing mutation rate from {self.mutation_rate:.2f} to {new_mutation_rate:.2f}"
+                f"Generation {generation}: Decreasing mutation rate from {self.mutation_rate:.3f} to {new_mutation_rate:.3f}"
             )
             self.mutation_rate = new_mutation_rate
 
-        # Adaptive Crossover
+        # Crossover rate adjustment
         if avg_diversity < diversity_low_threshold:
-            new_crossover_rate = min(0.9, self.crossover_rate * 1.05)
+            new_crossover_rate = max(0.1, self.crossover_rate * 0.95)
```
riesentoaster, line 77: Nice catch! Makes sense in my mind: If the population is not diverse enough, decrease crossover rate to prevent inbreeding.

But also: Did you talk to whoever implemented this first to check if there's a reason they did it the other way around?

src/fandango/evolution/adaptation.py
```diff
@@ -36,71 +41,90 @@ def update_parameters(
...
-                or new_max_repetition <= self.max_repetitions
-            ):
+            increment = math.ceil(
+                self.max_repetition_rate * self.current_max_repetition
+            )
+            increment = max(1, increment)
+            new_max_repetition = self.current_max_repetition + increment
+
+            if self.max_repetitions is not None:
+                new_max_repetition = min(new_max_repetition, self.max_repetitions)
```
riesentoaster, line 101: When would this ever happen?

PR #654, opened 2025-09-04
Move to CMake/skbuild build system

riesentoaster:


riesentoaster:
This is mostly vibe coded. It seems to work so far, but we should ensure that the CI build trial runs properly before we actually release this.

PR #656, opened 2025-09-10
Fix/test max nodes unsat

alex9849:
Fandango sometimes generated individuals that fulfilled the unsat (test_max_nodes_unsat) test. This here fixes that problem.


src/fandango/evolution/algorithm.py
```diff
@@ -248,21 +248,30 @@ def _perform_crossover(
                 child1, child2 = self.crossover_operator.crossover(
                     self.grammar, parent1, parent2
                 )
+                to_add = [child1, child2]
+                to_add = list(
+                    filter(lambda tree: tree.size() <= self.current_max_nodes, to_add)
+                )
```
riesentoaster: Why not just

```suggestion
                children = self.crossover_operator.crossover(
                    self.grammar, parent1, parent2
                )
                to_add = list(
                    filter(lambda tree: tree.size() <= self.current_max_nodes, children)
                )
```

tests/test_cli.py
```diff
@@ -302,13 +302,9 @@ def test_max_nodes_unsat(self):
...
-        err_pattern = r"""fandango:ERROR: Population did not converge to a perfect population
-fandango:ERROR: Only found (\d) perfect solutions, instead of the required 10"""
-        out_pattern = (
-            r"""(aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\.\d+\n)*"""
-        )
+        err_pattern = r"fandango:ERROR: Only found (0) perfect solutions, instead of the required 10"
         out, err, code = run_command(command)
-        self.assertRegex(out, out_pattern, f"out: {out}")
+        self.assertEqual(out, '', f"out: {out}")
         self.assertRegex(err, err_pattern, f"err: {err}")
```
riesentoaster: We don't need regex here anymore, I don't think:
```suggestion
        out, err, code = run_command(command)
        self.assertEqual(out, '', f"out: {out}")
        self.assertEqual(err, "fandango:ERROR: Only found 0 perfect solutions, instead of the required 10")
```

PR #709, opened 2025-10-01
Revert to old build system to re-enable beartype

riesentoaster:
beartype doesn't play nicely with scikit-build-core (see https://github.com/beartype/beartype/issues/556). Since we don't have to recompile that often (it essentially is only annoying because the CI runs a bit longer), I would rather have correct types than fast compile times.


src/fandango/beartype.py
```diff
@@ -3,6 +3,9 @@ def activate_beartype() -> None:
     from beartype import BeartypeConf
 
     skip_packages = (
+        "fandango.language.grammar.grammar",  # GrammarProcessor sometimes passes searches instead of NonTerminalNodes to Grammar.set_generator
+        "fandango.language.grammar.literal_generator",  # The above is then passed to the constructor of LiteralGenerator
+        "fandango.converters.dtd.DTDFandangoConverter",  # the type seems wrong
```
riesentoaster, line 8: A couple of things remain somewhat broken, but I'd rather merge now and fix those later. All of them are related to spec conversion.

src/fandango/constraints/base.py
```diff
@@ -142,10 +143,12 @@ def get_failing_nodes(
         return self.fitness(tree, scope, population, local_variables).failing_trees
 
     def __str__(self) -> str:
-        raise RuntimeError("Not implemented, use method specific to your usecase")
+        warnings.warn("Don't rely on this, use method specific to your usecase")
+        return self.format_as_spec()
```
riesentoaster, line 147: @alex9849 I have changed the `__repr__` and `__str__` fail-fast implementations with some that produce a result but will emit a warning. This way, we can just ignore them when we are debugging or messing around, but are still somewhat forced to build proper code that doesn't rely on these before we merge stuff.

tests/test_evaluation.py
```diff
@@ -1,21 +1,12 @@
...
-        stdout=subprocess.PIPE,
-        stderr=subprocess.PIPE,
-    )
-
-    print("STDOUT:\n", result.stdout.decode())
-    print("STDERR:\n", result.stderr.decode())
-
-    assert result.returncode == 0, "Script did not exit cleanly"
+    sys.setrecursionlimit(100000)
+    run_evaluation(time="1")
```
riesentoaster, line 12: @joszamama Is there a reason we didn't just call the evaluation directly but instead went through `subprocess.run`?

Also: I had to increase the recursion because beartype introduces an additional call for each call and that pushed it over the edge since we're producing rather large trees (and thus long call stacks) in some of the evals (specifically XML).

src/fandango/constraints/base.py
alex9849: Yes that sound great. :)

tests/test_evaluation.py
joszamama: IIRC, i had to manually modify the recursion limit for the depth of the fuzzer to reach more valid inputs.

PR #710, opened 2025-10-01
Fixes in main algorithm?

riesentoaster:
During a discussion about the algorithm, we found a couple of weird looking things. I'll add a couple of notes below. Are these actual bugs or am I missing something fundamental?

### DIFF
--- a/src/fandango/evolution/algorithm.py
+++ b/src/fandango/evolution/algorithm.py
@@ -247,6 +247,9 @@ class Fandango:
                 crossovers = self.crossover_operator.crossover(
                     self.grammar, parent1, parent2
                 )
+                if crossovers is None:
+                    return None
+
                 to_add = [
                     tree
                     for tree in crossovers
@@ -297,9 +300,6 @@ class Fandango:
                 except Exception as e:
                     LOGGER.error(f"Error during mutation: {e}")
                     print_exception(e, "Error during mutation")
-                    mutated_population.append(individual)
-            else:
-                mutated_population.append(individual)
         new_population.extend(mutated_population)
 
     def _perform_destruction(
@@ -391,11 +391,11 @@ class Fandango:
             new_population, unique_hashes = self._perform_selection()
 
             # Crossover
-            while (
-                len(new_population) < self.population_size
-                and random.random() < self.adaptive_tuner.crossover_rate
-            ):
-                yield from self._perform_crossover(new_population, unique_hashes)
+            for _ in range(self.population_size):
+                if len(new_population) >= self.population_size:
+                    break
+                if random.random() < self.adaptive_tuner.crossover_rate:
+                    yield from self._perform_crossover(new_population, unique_hashes)
 
             # Truncate if necessary
             if len(new_population) > self.population_size:
--- a/src/fandango/evolution/crossover.py
+++ b/src/fandango/evolution/crossover.py
@@ -1,5 +1,6 @@
 import random
 from abc import ABC, abstractmethod
+from typing import Optional
 
 from fandango.language.grammar.grammar import Grammar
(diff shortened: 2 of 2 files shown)


src/fandango/evolution/algorithm.py
```diff
@@ -297,9 +300,6 @@ def _perform_mutation(
                 except Exception as e:
                     LOGGER.error(f"Error during mutation: {e}")
                     print_exception(e, "Error during mutation")
-                    mutated_population.append(individual)
-            else:
-                mutated_population.append(individual)
```
riesentoaster, line 302: We shouldn't return unchanged elements here, right? This just adds duplicates to the population, which doesn't seem great.

src/fandango/evolution/crossover.py
```diff
@@ -9,19 +10,19 @@ class CrossoverOperator(ABC):
...
     def crossover(
         self, grammar: Grammar, parent1: DerivationTree, parent2: DerivationTree
-    ) -> tuple[DerivationTree, DerivationTree]:
+    ) -> Optional[tuple[DerivationTree, DerivationTree]]:
         symbols1 = parent1.get_non_terminal_symbols()
         symbols2 = parent2.get_non_terminal_symbols()
         common_symbols = symbols1.intersection(symbols2)
         if not common_symbols:
-            return parent1, parent2
+            return None
```
riesentoaster, line 25: Same thing here: Why return unchanged elements? We add them to the population and evaluate them, just wasting runtime.

src/fandango/evolution/algorithm.py
```diff
@@ -247,6 +247,9 @@ def _perform_crossover(
                 crossovers = self.crossover_operator.crossover(
                     self.grammar, parent1, parent2
                 )
+                if crossovers is None:
+                    return None
+
```
riesentoaster, line 252: This then just forwards how crossover doesn't return unchanged elements anymore.

src/fandango/evolution/algorithm.py
```diff
@@ -391,11 +391,11 @@ def _generate_simple(
...
-            while (
-                len(new_population) < self.population_size
-                and random.random() < self.adaptive_tuner.crossover_rate
-            ):
-                yield from self._perform_crossover(new_population, unique_hashes)
+            for _ in range(self.population_size):
+                if len(new_population) >= self.population_size:
+                    break
+                if random.random() < self.adaptive_tuner.crossover_rate:
+                    yield from self._perform_crossover(new_population, unique_hashes)
```
riesentoaster, line 398: And this was just calculated incorrectly we think.

This will lead to a weird distribution of how often crossover is actually executed, centering around very few times. By default, the crossover rate is 0.8. With a default population size of 100, even if the `new_population` were empty, this would likely only be executed a bit more than 3 times on average. If the population gets bad enough diversity (which it is unlikely to do considering we're barely doing crossover), the crossover rate is adjusted by the adaptive tuner to 0.9 — but if the diversity is ok, it gets reduced, down to 0.1.

This change would adjust it to run with a more expected rate.

riesentoaster:
One more question that came up: why do we only mutate the elite? And not the rest of the population? We should promote the elite to the next generation, yes. But give all individuals the chance to procreate through mutation/crossover.

src/fandango/evolution/crossover.py
joszamama: Yes, thank you for noticing!

joszamama:
> One more question that came up: why do we only mutate the elite? And not the rest of the population? We should promote the elite to the next generation, yes. But give all individuals the chance to procreate through mutation/crossover.

I have created an issue for this and I will fix it in a different branch.

PR #717, opened 2025-10-27
Fix/elite set mutation

joszamama:
DRAFT


src/fandango/evolution/algorithm.py
```diff
@@ -128,11 +130,14 @@ def __init__(
         self.crossover_operator = crossover_method
         self.mutation_method = mutation_method
 
+        # TODO: do this only if initial_population is provided.
```
riesentoaster, line 115: If you look at the function, it immediately returns if there is no initial population, so this is probably already ok

src/fandango/evolution/algorithm.py
```diff
@@ -128,11 +130,14 @@ def __init__(
...
         self.population = self._parse_and_deduplicate(population=initial_population)
         self._initial_solutions, self.evaluation = GeneratorWithReturn(
             self.evaluator.evaluate_population(self.population)
         ).collect()
 
+        # TODO: if no initial population, generate initial population here
```
riesentoaster, line 121: We at some point purposefully started lazy initializing it

src/fandango/evolution/algorithm.py
```diff
@@ -232,6 +241,9 @@ def _perform_crossover(
         :param new_population: The new population to perform crossover on.
         :param unique_hashes: The set of unique hashes of the individuals in the new population.
         """
+        # TODO: crossover should choose from non-elite set only
```
riesentoaster, line 206: Should it?

src/fandango/evolution/algorithm.py
```diff
@@ -232,6 +241,9 @@ def _perform_crossover(
...
+        # TODO: crossover % should mean: if %, cross a single individual, and remove both parents. else, copy random individual from previous gen. remove said individual from selection pool.
```
riesentoaster, line 207: You need to be very careful with this approach with your probabilities, as you don't always remove the same number of individuals from the population. It can work for sure, just keep this in mind so the param setting this probability still behaves intuitively.

src/fandango/evolution/algorithm.py
```diff
@@ -232,6 +241,9 @@ def _perform_crossover(
...
+        # TODO: since we produce 1 input per 2 parents, we fill the population until we reach the target size. (NOT DONE HERE, BUT IN generate_simple)
```
riesentoaster, line 208: Huh, so we suddenly generate from scratch again? No? Didn't we talk about how this isn't typical?

src/fandango/evolution/algorithm.py
```diff
@@ -284,6 +296,7 @@ def _perform_mutation(
 
         :param new_population: The new population to perform mutation on.
         """
+        # TODO: remove mutation pool, does not make sense. all population is subject to mutation based on %.
```
riesentoaster: This is an example of a significant change to the current behavior. We may want to release a new at least minor version after this.

src/fandango/evolution/algorithm.py
```diff
@@ -311,6 +324,7 @@ def _perform_destruction(
         :param new_population: The new population to perform destruction on.
         :return: The new population after destruction.
         """
+        # TODO: find most efficient way to destroy % of population at random.
```
riesentoaster, line 286: How about we remove destruction for now?

src/fandango/evolution/algorithm.py
```diff
@@ -369,6 +383,7 @@ def _generate_simple(
         :param max_generations: The maximum number of generations to generate. If None, the generation will run indefinitely.
         :return: A generator of DerivationTree objects, all of which are valid solutions to the grammar (or satisify the minimum fitness threshold).
         """
+        # TODO: since initial population was produced during init, we can remove this check.
```
riesentoaster, line 349: See other comment

src/fandango/evolution/algorithm.py
```diff
@@ -379,6 +394,7 @@ def _generate_simple(
         generation = 0
 
         while True:
+            # TODO: check the stop condition. if fitness is 1.0?
```
riesentoaster, line 359: Why would we stop generating then? Maybe I want more solutions?

src/fandango/evolution/algorithm.py
```diff
@@ -417,6 +440,7 @@ def _generate_simple(
                 self.population_size,
             )
 
+            # TODO: are we double evaluating here?
```
riesentoaster, line 404: We are. All over the place, sometimes nested. Check all the functions when `evaluate_individual` is passed. Maybe create a wrapper around a tree that contains the eval? This would make it a bit cleaner and would make it easier to reuse previous evals.

PR #724, opened 2025-11-03
Replace `argparse.FileType` with manual implementations

henryhchchc:
`argparse.FileType` is marked as [deprecated since Python 3.14](https://docs.python.org/3/library/argparse.html#argparse.FileType). We replace it with manual implementations.

Additional fix: 
- Close the fandango spec files after parsing.
- Add `types-setuptools` to enable mypy to type-check `setup.py`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ development = [
     "pre-commit>=4.2.0",
     "pygls>=2.0.0a4",           # language server
     "speedy-antlr-tool>=1.4.3",
-    "types-setuptools>=80.9.0", # enable mypy to check setup.py
+    "types-setuptools>=73.0.1", # enable mypy to check setup.py
 ]
 
 book = [


riesentoaster:
> > Would it make sense to require the same version for setuptools and types-setuptools?
> 
> Yeah I think so. Let me figure out which `setuptools` we are using.

From the top of `pyproject.toml`:

```toml
[build-system]
requires = ["setuptools>=73.0.1", "wheel"]
build-backend = "setuptools.build_meta"
```

henryhchchc:
> > > Would it make sense to require the same version for setuptools and types-setuptools?
> > 
> > 
> > Yeah I think so. Let me figure out which `setuptools` we are using.
> 
> From the top of `pyproject.toml`:
> 
> ```toml
> [build-system]
> requires = ["setuptools>=73.0.1", "wheel"]
> build-backend = "setuptools.build_meta"
> ```

Got it. Done.

PR #730, opened 2025-12-01
Exclude beartype 0.22.{6,7}

### LINKED ISSUE #729: Many tests are failing for unknown reasons (`eval_individual`)
The tests are passing two weeks ago but many of them are failing these few days. The errors originate from the following location:
```
src/fandango/evolution/population.py:89:

>           _fitness, failing_trees, suggestion = yield from eval_individual(individual)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E           TypeError: cannot unpack non-iterable NoneType object
```

However, none of the `return`s in `eval_individual` can return a `None`, so I suspect problem to be in some wrapper/decorator (if any). Given the tests passing two weeks ago, I think this might be caused by some breaking changes in our dependencies released in these two weeks.

Maybe we need some tools that can lock the dependencies. [uv](https://docs.astral.sh/uv/) is one worth a try.

henryhchchc:
0.22.6 breaks many tests

Fixes #729

### DIFF
--- a/src/fandango/language/grammar/nodes/alternative.py
+++ b/src/fandango/language/grammar/nodes/alternative.py
@@ -8,7 +8,7 @@ from fandango.language.grammar.nodes.node import Node, NodeType
 from fandango.language.tree import DerivationTree
 
 if TYPE_CHECKING:
-    import fandango
+    import fandango.language.grammar.node_visitors.node_visitor
 
 
 class Alternative(Node):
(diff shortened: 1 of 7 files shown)


henryhchchc:
`bear 0.22.8` is released so I updated the version spec to exclude problematic versions.

henryhchchc:
@Glinte That's weird but anyway I updated all the occurrences to make it consistent.

Issue #731, opened 2025-12-01
Dependency locking?


henryhchchc:
I think we need some mechanisms to prevent things broken by dependency updates (e.g., #729).

We have two options:

1. We use exact dependency versions in `pyproject.toml`. This is the easier way but may bring problems like dependency conflicts.
2. We keep using the relaxed `pyproject.toml` and introduce an extra `.lock` file. Tools like [uv](https://docs.astral.sh/uv/) support this but `pip lock` is still experimental.

What do you think?

Extra information:
- [Cargo.toml vs Cargo.lock](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html) for Rust but the concept can generalize to Python
- [Locking and Syncing - uv](https://docs.astral.sh/uv/concepts/projects/sync/)
- https://github.com/astral-sh/uv/issues/6271
- [`pip lock`](https://pip.pypa.io/en/stable/cli/pip_lock/)

Note: either way we can introduce @dependabot to automatically create PRs for dependency updates.

riesentoaster:
If we switched to `uv`, we would need to adjust all of the documentation and the CI workflows. It would also mean that if somebody does a `pip install` instead of going through `uv`, they may still receive incompatible versions, right?

What would be the downside of just using pinned versions in `pyproject.toml`? Disclaimer: I don't know very much about python packaging. 

Regardless, I really like the idea of dependabot PRs, they would then automatically run the CI, and we'd immediately know if a version is compatible.

PR #736, opened 2025-12-10
Lock dependencies with `uv`

henryhchchc:
Fixes: #731

### DIFF
--- a/.github/workflows/setup-fandango/action.yml
+++ b/.github/workflows/setup-fandango/action.yml
@@ -38,6 +38,7 @@ runs:
       uses: astral-sh/setup-uv@v7
       with:
         python-version: ${{ inputs.python-version }}
+        activate-environment: true
 
     - name: Install Python dependencies
       shell: bash
--- a/.github/workflows/benchmarks-main.yml
+++ b/.github/workflows/benchmarks-main.yml
@@ -30,7 +30,6 @@ jobs:
         env:
           FANDANGO_FORCE_SKIP_BEARTYPE: "1"
         run: |
-          source .venv/bin/activate
           bencher run \
           --project fandango-394f926-x0wdsa1zilqt3 \
           --token '${{ secrets.BENCHER_API_TOKEN }}' \
--- a/.github/workflows/benchmarks-prs.yml
+++ b/.github/workflows/benchmarks-prs.yml
@@ -23,7 +23,6 @@ jobs:
         env:
           FANDANGO_FORCE_SKIP_BEARTYPE: "1"
         run: |
-          source .venv/bin/activate
           pytest -n 0 --timeout 1200 --benchmark-json results.json --benchmark-enable --benchmark-only
       - name: Upload Benchmark Results
         uses: actions/upload-artifact@v4
--- a/.github/workflows/deploy-book.yml
+++ b/.github/workflows/deploy-book.yml
@@ -34,7 +34,6 @@ jobs:
       # Step 4: Locate Fandango and add to PATH
       - name: Locate Fandango and add to PATH
         run: |
-          source .venv/bin/activate
           which fandango
           echo "$(dirname $(which fandango))" >> $GITHUB_PATH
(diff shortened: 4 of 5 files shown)


henryhchchc:
> I mean, it's not pretty, but if it works, I don't see much that can go wrong, and it's documented, so 🤷‍♂️

Yeah but this is probably the best we can do – the warning originates from a dependency that hasn't been touched for six years :(

> What is still missing from this PR?

Documentation and polishing :)

henryhchchc:
I think that's it. If the tests are passing we are good to go.

riesentoaster:
image

This seems off. Can you take a look?

I think it broke in #724, at least that's the first PR where this step took significantly less long than before.

Also: The install Fandango step seems to take ~2mins longer than before. Is this correct or just a weird blip? Would you know a reason for that? And maybe a fix?

henryhchchc:
@riesentoaster `jupyter-book` updates from 1 to 2, which seem to change the CLI usage. I am pinning it to 1.x and see if it works.

riesentoaster:
Can you create an issue and assign it to Andreas, just so it is documented? He's the one responsible for the docs. Also: Do you know why this didn't fail the CI? We should set it up in a way where if the book fails, the CI run also fails.

Finally: Did you have a chance to look into the setup performance? Maybe a caching issue?

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
 requires = ["setuptools>=73.0.1", "wheel"]
-build-backend = "setuptools.build_meta"
+build-backend = "setuptools.build_meta:__legacy__"
```
riesentoaster: What's up with this?

.github/workflows/setup-fandango/action.yml
```diff
@@ -8,10 +8,6 @@ inputs:
   needs-sudo:
     description: "Whether to use sudo"
     required: true
-  target:
-    description: "Which dependency list to target"
-    required: false
-    default: "test"
```
riesentoaster, line 14: What is the reason for this change? I added this code initially to only install the required packages, to save CI time. How much longer does it take to always install all dependencies? Is it even noticeable? If not I'm all in favor of reducing this complexity.

riesentoaster (changes requested):

pyproject.toml
henryhchchc: This is to make the C++ parser build work with `uv sync` (and also `pip install`). Without this we have to add an extra `uv pip install`.

riesentoaster: And I assume the latter is slower?

What's the underlying difference between the two invocations of uv?

riesentoaster: I have been somewhat tempted to try to take the C++ parser out of the repo and push it into its own thing. Would make the entire build system thing so much easier probably. We should look into that at some point, we've talked about splitting the API and the CLI anyways. But none of that should happen in this PR.

.github/workflows/setup-fandango/action.yml
henryhchchc: This is for better caching – in this way `uv` can use the same cache across steps. For example, my intention is to make steps like tests benefit from steps like `mypy check`. What do you think?

henryhchchc:
> this is a decent quality of life improvement for Fandango development

Lol one thing that I found annoying is the compilation of C++ parsers, which uses a single core (while my machine has ten). Do you have an idea on how to parallelize it?

.github/workflows/setup-fandango/action.yml
riesentoaster: I am not entirely sure how caching works, but I think that it may have worked previously as well.

But then also: I don't think it adds too much runtime, at least at a cursory check. So maybe the code simplicity should be prioritized. 

I'm fine leaving it as is.

riesentoaster (approved):
Good idea. Thank you for all your work! Let's get this merged!

PR #737, opened 2025-12-15
Merge protocol guidance

alex9849:
Added coverage guidance system for IO grammars


src/fandango/evolution/algorithm.py
```diff
@@ -89,7 +92,10 @@ def __init__(
         self.tournament_size = tournament_size
         self.warnings_are_errors = warnings_are_errors
         self.best_effort = best_effort
+        self.current_max_nodes = max_nodes
+        self.diversity_k = diversity_k
```
riesentoaster, line 98: We shouldn't have information redundancy such as this. `AdaptiveTuner` owns this information, and it should be the only source of truth for this. If you need to be able to reset it, make `AdaptiveTuner` be able to do this. And then access the information from there directly to prevent out-of-sync errors.

src/fandango/evolution/population.py
```diff
@@ -26,22 +27,21 @@ def _generate_population_entry(self, max_nodes: int) -> DerivationTree:
         return self._grammar.fuzz(self._start_symbol, max_nodes)
 
     @staticmethod
+    def _generate_population_hashes(
+        current_population: list[DerivationTree],
+    ) -> set[int]:
+        return {hash(ind) for ind in current_population}
+
     def add_unique_individual(
+        self,
```
riesentoaster: Why is this no longer static?

src/fandango/evolution/algorithm.py
```diff
@@ -477,111 +490,211 @@ def _generate_io(
...
+        assert isinstance(self.evaluator, IoEvaluator)
 
         while True:
-            forecast = forecaster.predict(history_tree)
+            self.packet_selector.compute(history_tree, self.past_io_derivations)
+            LOGGER.info(
+                f"Current coverage: {self.packet_selector.coverage_percent():.2f}%"
+            )
+            self.evaluator.start_next_message(
+                [history_tree] + list(self.past_io_derivations)
```
riesentoaster: Nitpick: This call to `list` is unnecessary

src/fandango/evolution/algorithm.py
```diff
@@ -477,111 +490,211 @@ def _generate_io(
         spec_env_global, _ = self.grammar.get_spec_env()
         io_instance: FandangoIO = spec_env_global["FandangoIO"].instance()
         history_tree: DerivationTree = random.choice(self.population)
-        forecaster = PacketForecaster(self.grammar)
+        self.packet_selector = PacketSelector(
```
riesentoaster, line 496: This function is way too complicated now. Please refactor in a way where it remains understandable. Think about where certain logic belongs as well.

IMHO this class should really only contain calls for each conceptual step in the evolutionary algorithm, such that I can read it and understand what is happening even if I don't know the Fandango internals, and sub-steps should either be moved to methods or (at least for some of these probably) to other classes.

src/fandango/evolution/population.py
```diff
@@ -154,3 +168,17 @@ def _generate_population_entry(self, max_nodes: int) -> DerivationTree:
...
+        candidate: DerivationTree,
+        unique_set: set[int],
+    ) -> bool:
+        new_hashes = self._generate_population_hashes([candidate])
+        if len(new_hashes.intersection(unique_set)) == 0:
+            # If the candidate has is new, we can add it to the population
+            unique_set.update(new_hashes)
+            population.append(candidate)
+            return True
+        return False
```
riesentoaster: This is duplicate code

src/fandango/language/grammar/nodes/node.py
```diff
@@ -57,6 +58,17 @@ def is_terminal(self) -> bool:
     def is_nonterminal(self) -> bool:
         return self.node_type == NodeType.NON_TERMINAL
 
+    @abc.abstractmethod
+    def to_symbol(self) -> Symbol:
```
riesentoaster, line 62: Document please

src/fandango/language/grammar/nodes/node.py
```diff
@@ -57,6 +58,17 @@ def is_terminal(self) -> bool:
...
+        raise NotImplementedError("to_symbol method not implemented")
+
+    @property
+    def is_controlflow(self) -> bool:
```
riesentoaster, line 70: Document please

src/fandango/language/grammar/nodes/node.py
```diff
@@ -77,10 +89,22 @@ def accept(
     ) -> "fandango.language.grammar.node_visitors.node_visitor.ResultType":
         raise NotImplementedError("accept method not implemented")
 
-    def msg_parties(self, *, include_recipients: bool = False) -> set[str]:
+    def msg_parties(
```
riesentoaster, line 100: Document please

I'll stop these comments here, but please look at all functions in this class, in particular those that have parameters that aren't entirely obvious.

tcp.fan
riesentoaster, line 1: This most likely shouldn't be here

docs/Contributing.md
```diff
@@ -77,9 +77,11 @@ To install additional dependencies for unit tests, code checks, running the eval
 $ python -m pip install -e ".[test,development,evaluation,book]"
 ```
 
-```{info}
-If subsequent tests fails due to dependency breakage, use `uv` to create the virtual environment with `uv.lock`, which provides a snapshot of the dependencies that are tested on our CI.
-`uv sync --locked --all-extras && uv pip install -e ".[test,development,evaluation,book]" --no-deps`
+```{tip}
+If subsequent tests fails due to dependency breakage, use `uv` to create the virtual environment with `uv.lock`, which provides a snapshot of the dependencies that are tested on our CI:
+````shell
+$ uv sync --locked --all-extras && uv pip install -e ".[test,development,evaluation,book]" --no-deps
```
henryhchchc:
```suggestion
$ uv sync --locked --all-extras
```

The second command is no-longer required after #739

alex9849:
Fixes #748 
Added self.grammar.populate_sources(tree) to prase-function in api.py

src/fandango/language/symbols/symbol.py
```diff
@@ -63,13 +64,17 @@ def format_as_spec(self) -> str:
...
 
     def __str__(self) -> str:
-        warnings.warn(
-            f"Don't rely on the __str__ impl on {self.__class__.__name__}, use method specific to your usecase. Report this as a bug if this is called from within Fandango."
-        )
+        if not TYPE_CHECKING:
+            pass
+            # warnings.warn(
+            #    f"Don't rely on the __str__ impl on {self.__class__.__name__}, use method specific to your usecase. Report this as a bug if this is called from within Fandango."
+            # )
```
riesentoaster, line 71: Why did we disable these warnings? I don't think we should do this. And we shouldn't return the `format_as_spec` representation here — users should think about what exact representation they want.

src/fandango/language/tree_value.py
```diff
@@ -676,7 +676,9 @@ def __hash__(self) -> int:
 
     def __repr__(self) -> str:
         if self._trailing_bits:
-            return f"{self._value!r} + bits: {''.join(str(bit) for bit in self._trailing_bits)}"
+            # When printing a Fandango spec, this breaks parsing - AZ
+            # return f"{self._value!r} + bits: {''.join(str(bit) for bit in self._trailing_bits)}"
+            return "".join(str(bit) for bit in self._trailing_bits)
```
riesentoaster, line 681: We shouldn't ignore non-trailing bits data. I do see how if a `TreeValue` only contains trailing bits, it makes more sense to only print trailing bits. But if there is other stuff in there (which can absolutely happen!), we shouldn't ignore it.

andreas-zeller: This is called when (pretty-)printing a Fandango spec; so the word "bits:" breaks syntax (and consequent parsing). Is there a way a `TreeValue` can have a non-bit value _and_ bits? If we don't assume this, maybe an assertion would be in order here.

riesentoaster: Yes it can contain other data. Imagine you have a spec that contains both first some bytes and then some bits. When we serialize its trees, you'll get `TreeValue`s that have bytes and then some bits. We can't immediately convert to bytes since we may not have enough bits.

I guess we could convert everything to bits when `__repr__` is called. That may however lead to other problems, not sure.

PR #739, opened 2025-12-17
Parallel C++ Compilation with CMake

henryhchchc:
As mentioned by @riesentoaster 

> There are about 10 PRs in Fandango's history of me trying to accomplish this. I have tried about 4 different build systems. It was working at some point but then I noticed that beartype broke so we reverted it again.
> 
> I seriously dislike python build systems to put it mildly. But also: I don't think the parsers change all that often, so reinstalling them is probably not necessary locally unless you're changing the actual meta-grammar. And we release prebuilt binaries for downstream use. It's mostly a CI bottle neck but there we only have 4 cores, so it'll save like 1.5min.
> 
> If we want to speed this up, we should probably
> 
> * Split the parser into a different package that wouldn't need to be rebuilt on each push, but we could rely on prebuilt binaries
> * Look at some of the slower unit tests and replace them if more precise ones (we have a lot of tests that run a lot of code to test one detail)
> * Improve Fandango's general performance lol
> 
> Maybe also: rebuild the compiled parser in rust, cargo is always parallelized :D

This PR introduces CMake for parallel compilation of the C++ parser.

- [x] Awaiting: #736

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,9 @@ project(${SKBUILD_PROJECT_NAME}
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+# Enable Link Time Optimization
+set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
+
 # Find Python (use Module development to avoid requiring libpython on manylinux)
 find_package(Python3 COMPONENTS Interpreter Development.Module REQUIRED)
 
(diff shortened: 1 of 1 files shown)


henryhchchc:
After digging into the source code of `setuptool`, I found it cannot compile source files in parallel (like `make -J`). 

https://github.com/pypa/setuptools/blob/d198e86f57231e83de87975c5c82bc40c196da79/setuptools/_distutils/compilers/C/base.py#L650-L655

It can build multiple extensions in parallel, though.

https://github.com/pypa/setuptools/blob/d198e86f57231e83de87975c5c82bc40c196da79/setuptools/_distutils/command/build_ext.py#L486-L505

henryhchchc:
Let me try again with `scikit-build-core`. I will start with e5c185002012ae471a291f56d91bcc706807a56b.

riesentoaster:
For future reference: This PR reverted scikit-based parallellism last time: #709

But yeah, beartype seems to work now! https://github.com/fandango-fuzzer/fandango/actions/runs/20305074912/job/58320185284

I guess we also have a test that tries to check if beartype is active, although I wouldn't trust this bodge 100%: https://github.com/fandango-fuzzer/fandango/blob/413b5f7ef1a7c887d1379a4bd0f9a8e2154d8c25/tests/test_meta.py#L13-L17

Thank you so much for trying again!

One thing though: The entire build system (e.g. CMakeLists.txt) seems rather complicated. Would it be possible to simplify them a bit to make them a bit less horrible to maintain? Unless you know what you're doing and are willing to take over the responsibility for that — I know that we don't really have someone with experience with these build systems and then I guess it makes sense to keep things as simple and stupid as possible. Maybe also check the impl from last time?

PR #764, opened 2026-01-13
Add greybox support

leonbett:
Depends on #783


.github/workflows/python-tests.yml
```diff
@@ -25,6 +25,17 @@ jobs:
...
+      - name: Set path for macOS
+        if: matrix.os == 'macos-latest'
+        run: |
+          echo "Adding Homebrew LLVM@18 to PATH"
+          echo "/opt/homebrew/opt/llvm@18/bin" >> "$GITHUB_PATH"
+
+      - name: Install fcc
+        if: matrix.os != 'windows-latest'
+        run: |
+          make fcc
```
riesentoaster: Could this be a dependency of the existing `make test-tools` target?

src/fandango/cli/parser.py
```diff
@@ -456,6 +456,20 @@ def _populate_fuzz_parser(parser: argparse.ArgumentParser) -> None:
...
+        default="lambda t: False",
+        help='stop criterion to be used. This is a lambda function which is run on every new solution. Example: `lambda t: t.to_string().startswith("abc")`',
+    )
+    parser.add_argument(
+        "--stop-after-seconds",
+        type=int,
+        dest="stop_after_seconds",
+        default=None,
+        help="Stop after a given number of seconds. Example: `--stop-after-seconds 60`",
+    )
```
riesentoaster, line 482: These are very cool, but probably independent of the coverage guidance, right? In that case, do we want to consider stabilizing these features and document them?

src/fandango/evolution/algorithm.py
```diff
@@ -90,6 +104,21 @@ def __init__(
         self.warnings_are_errors = warnings_are_errors
         self.best_effort = best_effort
         self.remote_response_timeout = 15.0
+        self.use_fcc = use_fcc
+        self.sa = None
+        self.da = None
+        self.put = put
+        self.put_args = put_args
```
riesentoaster: These class fields are getting really quite crowded, there are so many of them. Would it make sense to extract the fcc specific logic into its own class? Just to keep the main algorithm class smaller and thus more understandable/maintainable?

src/fandango/evolution/algorithm.py
```diff
@@ -90,6 +104,21 @@ def __init__(
...
+        self.experiment_start_time = time.time()
+        self.stop_after_seconds = stop_after_seconds
+        if self.use_fcc:
+            assert self.put
+            root_directory: str = os.path.dirname(self.put)
+            cfg_directory: str = os.path.join(root_directory, ".cfg/")
+            bb_to_dbg_json: str = os.path.join(root_directory, ".bb_to_dbg.json")
+            LOGGER.info("Running fcc static analysis...")
+            self.sa = StaticAnalysis(cfg_directory, bb_to_dbg_json)
+            self.da = DynamicAnalysis(self.sa, root_directory, self.put, self.put_args)
```
riesentoaster: All of this could then go into a `self.fcc_helper.init()` call or whatever.

src/fandango/evolution/algorithm.py
```diff
@@ -378,91 +421,109 @@ def _generate_simple(
...
+            while True:
+                if max_generations is not None and generation >= max_generations:
+                    break
+                generation += 1
 
-            avg_fitness = sum(e[1] for e in self.evaluation) / self.population_size
+                # gc.collect()
 
-            LOGGER.info(f"Generation {generation} - Average Fitness: {avg_fitness:.2f}")
+                self.throw_if_time_limit_reached()
```
riesentoaster: Why do we throw if the timeout is reached, and not just break/stop the generator like with the other stopping criteria?

src/fandango/evolution/evaluation.py
```diff
@@ -25,6 +28,9 @@ def __init__(
         diversity_k: int,
         diversity_weight: float,
         warnings_are_errors: bool = False,
+        sa: Optional[StaticAnalysis] = None,
+        da: Optional[DynamicAnalysis] = None,
```
riesentoaster: Please use non-minimized variable names

src/fandango/experimental/execution/static_analysis.py
riesentoaster, line 1: Would it make sense to split this file into multiple files? It's rather long.

Also: Why are some methods in here snake case and some pascal case?

riesentoaster (changes requested):
I love how little change is actually needed for this to work. And I love the number/depth of tests you add. 

I do have some initial comments, will do a second, more thorough pass once these bigger things are cleared up/we agree not to change them.

src/fandango/evolution/algorithm.py
leonbett: Good point!

.github/workflows/python-tests.yml
leonbett: Yes, for the "install fcc" part.
The "macOS" part is a bit tricky, because it's CI-specific, and a requirement for fcc.
So if we install fcc as part of test-tools, LLVM is not yet set up on macOS.

src/fandango/evolution/algorithm.py
leonbett: Done

src/fandango/evolution/algorithm.py
leonbett: To be honest, I don't remember why I implemented it that way back when I first wrote the code.
I've now changed it to work without exceptions, and it seems to work still. Thanks!

src/fandango/experimental/execution/static_analysis.py
leonbett: Good idea! Done.

snake case is for internal functions, pascal case is for user-facing functions that can be used as part of a constraint.

src/fandango/cli/parser.py
leonbett: Sure, these could be used for other experiments, too. I'll add a separate PR.

.github/workflows/python-tests.yml
riesentoaster: I would suggesting merging the make commands where possible.

For the LLVM PATH issue: Is that also a problem on a local macOS installation? If yes, this should be part of the Makefile as well. And we should document this somewhere.

leonbett: I added a check in the Makefile.

src/fandango/api.py
```diff
@@ -407,6 +410,7 @@ def fuzz(
                 desired_solutions,
                 max_generations,
                 infinite,
+                settings["use_fcc"] if "use_fcc" in settings else False,
```
riesentoaster:
```suggestion
                settings.get("use_fcc", False)
```

src/fandango/api.py
```diff
@@ -407,6 +410,7 @@ def fuzz(
                 desired_solutions,
                 max_generations,
                 infinite,
+                settings.get("use_fcc", False)
```
riesentoaster:
```suggestion
                settings.get("use_fcc", False),
```

src/fandango/cli/utils.py
```diff
@@ -97,6 +97,11 @@ def make_fandango_settings(
         # previously is a str, we eval it into a function
         settings["stop_criterion"] = eval(args.stop_criterion)
     _copy_setting(args, settings, "stop_after_seconds")
+    _copy_setting(args, settings, "use_fcc")
+
+    if "use_fcc" in settings:
```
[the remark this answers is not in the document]
riesentoaster:
```suggestion
    if settings.get("use_fcc", False):
```

tests/test_execution_feedback.py
```diff
@@ -0,0 +1,254 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls.put_dir = "tests/resources/execution_feedback/reachability_1"
+
+    def test_fcc_installed(self):
+        # Check that fcc is installed
+        try:
+            subprocess.check_output([FCC_PATH, "--version"])
+        except FileNotFoundError as e:
+            self.fail(f"Command not found: {e}")
```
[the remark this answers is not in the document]
riesentoaster: That's just wrong, `make fcc-local` puts it exactly at `FCC_PATH`, and installing it non-locally still leaves it there.

PR #782, opened 2026-01-21
Run make web in CI/Fix docs failing silently in the CI

riesentoaster:


.github/workflows/build-docs.yml
```diff
@@ -23,27 +23,16 @@ jobs:
...
-          echo "$(dirname $(which fandango))" >> $GITHUB_PATH
-          fandango --help
-
-      # Step 5: Build the documentation
+      
+      # Step 3: Build the documentation
       - name: Build the book
         run: |
+          sudo make system-dev-tools
           source .venv/bin/activate
```
henryhchchc:
```suggestion
```
This is not needed as `setup-fandango` will activate the `.venv` for us.

.github/workflows/build-docs.yml
```diff
@@ -23,27 +23,16 @@ jobs:
...
-          make html
+          bash -c "source .venv/bin/activate && make web"
```
henryhchchc:
```suggestion
          make web
```
- Activating the venv is also unnecessary here
- We can make the entire step to use `bash` (see [this suggestion](https://github.com/fandango-fuzzer/fandango/pull/782/changes#r2715237932))

.github/workflows/build-docs.yml
```diff
@@ -23,27 +23,16 @@ jobs:
...
-        run: |
-          which fandango
-          echo "$(dirname $(which fandango))" >> $GITHUB_PATH
-          fandango --help
-
-      # Step 5: Build the documentation
+      
+      # Step 3: Build the documentation
       - name: Build the book
         run: |
```
henryhchchc:
```suggestion
        shell: bash
        run: |  
```

We can set the entire step to run in `bash`.

henryhchchc (commented):
LGTM (with a few nitpicks)!

henryhchchc (approved):

henryhchchc:
> But in the CI it apparently just runs through. What else could it be?

That's weird. Is there anything in the makefile suppressing this error?

andreas-zeller:
You can also integrate an `AssertionError` check right into the CI. If there are any `AssertionError` outputs in the docs, then this should have an exit code of 1:

grep -l AssertionError docs/_build/html/*.html; test $? != 0

henryhchchc:
Great! Looks like it fails with `AssertionError`s now.

PR #783, opened 2026-01-21
Add stop options

leonbett:
These are useful for setting up Fandango experiments, as explained in the comments.


src/fandango/evolution/algorithm.py
```diff
@@ -69,6 +69,8 @@ def __init__(
         max_nodes: int = 200,
         max_nodes_rate: float = 0.5,
         profiling: bool = False,
+        stop_criterion: str = "lambda t: False",
```
riesentoaster: The arg here shouldn't be a string to be `eval`ed, it should be the optional parsed function. And the same applies to the outside interface. And if the arg isn't passed in the command line, this should be an empty optional and not a constant function, just like in downstream calls.

riesentoaster (changes requested):
We should probably also add this to the docs somewhere. @andreas-zeller what do you think? Suggestions?

PR #791, opened 2026-01-22
Fix validation bug

### LINKED ISSUE #777: Validation fails when generating binaries
The command

```shell
$ fandango fuzz -n 1 -f docs/binary-pack.fan --validate
```

produces `FandangoError: 1 error(s) during validation`. This is a regression vs. 1.0.

The diff points to _string_ types being produced, and _byte_ types being parsed.

```
FandangoError: '/var/folders/n2/xd9445p97rb3xh7m1dfx8_4h0006ts/T/tmphjiloech/fandango-0000.txt': parsed tree does not match original
  *** original

--- parsed

***************

*** 1,22 ****

  <start> ::= <field>
!   <field> ::= <length> <content>  # 3461849
!     <length> ::= <byte> <byte>  # 
        <byte> ::= <_byte>
!         <_byte> ::= '\x07'  # Position 0x0000 (0)
        <byte> ::= <_byte>
!         <_byte> ::= '\x00'  # Position 0x0001 (1)
!     <content> ::= <digit> <digit> <digit> <digit> <digit> <digit> <digit>  # 3461849
        <digit> ::= <_digit>
!         <_digit> ::= '3'  # Position 0x0002 (2)
        <digit> ::= <_digit>
!         <_digit> ::= '4'  # Position 0x0003 (3)
        <digit> ::= <_digit>
!         <_digit> ::= '6'  # Position 0x0004 (4)
        <digit> ::= <_digit>
!         <_digit> ::= '1'  # Position 0x0005 (5)
        <digit> ::= <_digit>
!         <_digit> ::= '8'  # Position 0x0006 (6)
        <digit> ::= <_digit>
!         <_digit> ::= '4'  # Position 0x0007 (7)
        <digit> ::= <_digit>
!         <_digit> ::= '9'  # Position 0x0008 (8)
--- 1,22 ----

  <start> ::= <field>
!   <field> ::= <length> <content>  # b'\x07\x003461849'
!     <length> ::= <byte> <byte>  # b'\x07\x00'
        <byte> ::= <_byte>
!         <_byte> ::= b'\x07'  # Position 0x0000 (0)
        <byte> ::= <_byte>
!         <_byte> ::= b'\x00'  # Position 0x0001 (1)
!     <content> ::= <digit> <digit> <digit> <digit> <digit> <digit> <digit>  # b'3461849'
        <digit> ::= <_digit>
!         <_digit> ::= b'3'  # Position 0x0002 (2)
        <digit> ::= <_digit>
!         <_digit> ::= b'4'  # Position 0x0003 (3)
        <digit> ::= <_digit>
!         <_digit> ::= b'6'  # Position 0x0004 (4)
        <digit> ::= <_digit>
!         <_digit> ::= b'1'  # Position 0x0005 (5)
        <digit> ::= <_digit>
!         <_digit> ::= b'8'  # Position 0x0006 (6)
        <digit> ::= <_digit>
!         <_digit> ::= b'4'  # Position 0x0007 (7)
        <digit> ::= <_digit>
!         <_digit> ::= b'9'  # Position 0x0008 (8)
FandangoError: 1 error(s) during validation
```

riesentoaster:
Closes #777

### DIFF
--- a/docs/binary-pack.fan
+++ b/docs/binary-pack.fan
@@ -7,4 +7,4 @@ where <length> == uint16(len(<content>))
 from struct import pack
 
 def uint16(n: int) -> str:
-    return pack('<H', n).decode('iso8859-1').encode()  # convert to bytes
+    return pack('<H', n) # convert to bytes


docs/binary-pack.fan
```diff
@@ -7,4 +7,4 @@ where <length> == uint16(len(<content>))
 from struct import pack
 
 def uint16(n: int) -> str:
-    return pack('<H', n).decode('iso8859-1')  # convert to string
\ No newline at end of file
+    return pack('<H', n) # convert to bytes
```
riesentoaster, line 10: The problem here is that we are very lenient with types. This constraint previously compared a string to a part of the tree (`<length>`). This nonterminal thus had to be linearized to a string. When we later linearize the entire tree, we need to merge a part of the tree linearized as bytes with is cached as a string. And at the moment, strings always win out in this conversion. Therefore, the entire tree was linearized to string.

It was then force converted to bytes for writing to a file again, because there we use a heuristic to check if we should open the output file in text or binary mode, and call `to_bytes()` or `to_string()` respectively (the former here).

This means the representation on disk is correct, but the comparison between the re-parsed file and the initial output doesn't know which way to convert the internal representation, and thus is strict about how both sides need to have the same type for the comparison to succeed and thus fail the validation.

We can easily fix this by preventing the initial problem with this spec and convert the length to bytes instead of strings, but we may want to think about a more general solution, since this will hide all sorts of potential issues. We could always emit a warning if `TreeValues` with internal representation of bytes and strings are merged, but that may also trigger problems where things are fine.

PR #792, opened 2026-01-24
Report fandango upgrades when available

andreas-zeller:
This pull request adds a message when Fandango updates are available on PyPi:

```fandango
$ fandango
📦 Update available for 'fandango-fuzzer': 1.0.5a0 → 1.0.6. See https://pypi.org/project/fandango-fuzzer/1.0.6/
...
```

The check
* runs before the actual command (say, `fuzz`) is executed
* runs only if the output is a terminal (i.e., we have an interactive session)
* outputs to standard error (and thus does not interfere with regular output)
* is suppressed when `CI` or `FANDANGO_DISABLE_UPDATE_CHECK` env variables are set
* does not check for updates in any dependent packages
* is performed only once a week, but `fandango version` explicitly triggers it.

### DIFF
--- a/src/fandango/cli/commands.py
+++ b/src/fandango/cli/commands.py
@@ -431,8 +431,9 @@ def copyright_command(args: argparse.Namespace) -> None:
 # `version_command()` is called from `shell_command()`
 # with `skip_update_check=True`, so we only check for updates when
 # the user explicitly requests the version via command line.
-def version_command(args: argparse.Namespace, *,
-                    skip_update_check: bool = False) -> None:
+def version_command(
+    args: argparse.Namespace, *, skip_update_check: bool = False
+) -> None:
     if sys.stdout.isatty():
         version_line = f"💃 {styles.color.ansi256(styles.rgbToAnsi256(128, 0, 0))}Fandango{styles.color.close} {fandango.version()}"
     else:
--- a/src/fandango/cli/commands.py
+++ b/src/fandango/cli/commands.py
@@ -428,12 +428,15 @@ def copyright_command(args: argparse.Namespace) -> None:
     print("All rights reserved.")
 
 
-# `version_command()` is called from `shell_command()`
-# with `skip_update_check=True`, so we only check for updates when
-# the user explicitly requests the version via command line.
 def version_command(
     args: argparse.Namespace, *, skip_update_check: bool = False
 ) -> None:
+    """
+    Show Fandango version and check for updates.
+    :param: skip_update_check - if True, do not force-check for updates
+    This is set when called from `shell_command()`, which reports the version.
+    """
+
     if sys.stdout.isatty():
         version_line = f"💃 {styles.color.ansi256(styles.rgbToAnsi256(128, 0, 0))}Fandango{styles.color.close} {fandango.version()}"
     else:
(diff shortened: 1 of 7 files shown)


src/fandango/cli/upgrade.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    If `package` has an update available on PyPI, print a notification to stderr.
+
+    :param package_name: package name
+    :param cache_dir: Where to store the cache file (default: ~/.cache)
+    :param check_now: Whether to check for updates immediately, bypassing the rate limit
+    :return: True if an update notification was printed, False otherwise
+    """
+    if not cache_dir:
+        cache_dir = Path.home() / ".cache"
```
riesentoaster, line 42: Does it make sense to store this here? Don't we have fandango files stored in other locations on the file systems? We shouldn't spread them around everywhere.

src/fandango/cli/upgrade.py
```diff
@@ -0,0 +1,115 @@
...
+from fandango.language.parse.cache import get_cache_dir
+from fandango.cli.parser import terminal_link
+from fandango import DISTRIBUTION_NAME
+from fandango.logger import LOGGER
+
+# How often to check for updates (in seconds)
+CHECK_INTERVAL_SECONDS = 7 * 24 * 60 * 60  # one week
+
+# Network timeout for PyPI requests
+TIMEOUT = 5  # seconds
```
riesentoaster, line 22: This will get noticed by users and confuse them ("Why does Fandango sometimes take so long to start?"). Do we want to print a message to stderr when we do this check? This could also contain the information about how to disable these checks.

Otherwise it is really hard to detect that this is even happening and how to disable it. Imagine people doing benchmarks between fuzzers etc.

src/fandango/cli/commands.py
```diff
@@ -427,12 +428,19 @@ def copyright_command(args: argparse.Namespace) -> None:
     print("All rights reserved.")
 
 
-def version_command(args: argparse.Namespace) -> None:
+# `version_command()` is called from `shell_command()`
+# with `skip_update_check=True`, so we only check for updates when
+# the user explicitly requests the version via command line.
```
riesentoaster: Nitpick: Any reason to not have this as a docstring?

riesentoaster (changes requested):

src/fandango/cli/upgrade.py
andreas-zeller: This is a generic function that can also be used outside of Fandango (and then, it uses the `~/.cache` default). Within Fandango, this function gets passed the [standard cache location](https://fandango-fuzzer.github.io/Including.html#where-does-fandango-store-parsed-files) to store time stamps.

src/fandango/cli/upgrade.py
andreas-zeller: I'm setting things up so Fandango issues a message

```
📦  Checking for Fandango updates... (set FANDANGO_DISABLE_UPDATE_CHECK=1 to disable)
```

while the check is running.

riesentoaster: Perfect, thank you!

src/fandango/cli/upgrade.py
riesentoaster: Not sure why we would need this to be generic, but sure.

riesentoaster (approved):
Thank you for the changes and the PR, I really like this!

henryhchchc:
The docs build is fixed in #737 so let's merge this!

PR #802, opened 2026-01-30
Fix: ANTLR undefined

andreas-zeller:
The Fandango `Makefile` does not define a value for ANTLR (should be `antlr`). This prevents development builds.


riesentoaster (approved):
How did this previously work?!

riesentoaster:
(This was already fixed in #737 I think?)

PR #817, opened 2026-02-23
Raise error if output file/dir exists

### LINKED ISSUE #815: Fandango fuzz output `-o` appends to existing input
When using `-o` to direct output to a file FILE, the output is _appended_ to the file.

After

```shell
$ rm out.txt
$ fandango fuzz -f persons.fan -o out.txt -n1
$ fandango fuzz -f persons.fan -o out.txt -n1
```

the file `out.txt` contains two instances of Fandango output. This is surprising; using `-o` should _overwrite_ existing content, not append to it. This is particularly nasty when generating (single) binary inputs, which typically become invalid when concatenating multiple instances.

If users want to append to an existing file, they should redirect stdout to the file using`>>`.

riesentoaster:
Fixes #815

### DIFF
--- a/src/fandango/cli/commands.py
+++ b/src/fandango/cli/commands.py
@@ -188,7 +188,7 @@ def fuzz_command(args: argparse.Namespace) -> None:
         out_dir = Path(args.directory)
         if out_dir.exists() and (not out_dir.is_dir() or len(os.listdir(out_dir)) > 0):
             raise FandangoError(
-                f"Output directory {out_dir} already exists as a non-file or is not empty"
+                f"Output directory {out_dir} is not a directory or is not empty"
             )
 
     def solutions_callback(sol: DerivationTree, i: int) -> None:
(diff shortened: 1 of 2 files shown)


src/fandango/cli/commands.py
```diff
@@ -178,6 +178,19 @@ def fuzz_command(args: argparse.Namespace) -> None:
...
+        p = Path(args.output)
+        if p.exists():
+            LOGGER.info(f"Removing existing output file {p}")
+            p.unlink()
+
+    if getattr(args, "directory", None) is not None:
+        out_dir = Path(args.directory)
+        if out_dir.exists() and (not out_dir.is_dir() or len(os.listdir(out_dir)) > 0):
+            raise FandangoError(
+                f"Output directory {out_dir} already exists as a non-file or is not empty"
```
andreas-zeller: Change this from "... exists as a non-file" to "Output {out_dir} is not a directory or is not empty". Thanks!

riesentoaster: Changed it to `<out_dir> is not a directory or is not empty` — you're right, this is much clearer.

PR #823, opened 2026-02-25
Io bugfixes

alex9849:
Add bugfixes from evaluation and documentation updates

### DIFF
--- a/src/fandango/io/__init__.py
+++ b/src/fandango/io/__init__.py
@@ -239,7 +239,7 @@ class ProtocolImplementation(ABC):
         self._party_instance = party_instance
 
     @property
-    def io_instance(self) -> FandangoIO:
+    def io_instance(self) -> "FandangoIO":
         return self._party_instance.io_instance
 
     def send(


src/fandango/io/__init__.py
```diff
@@ -160,12 +160,16 @@ def send(
...
-    def receive(self, message: str | bytes, sender: Optional[str]) -> None:
+    def receive(self, message: str | bytes | None, sender: Optional[str]) -> None:
         """
         Called when a message has been received by this party.
         :param message: The content of the message.
         :param sender: The sender of the message.
         """
+        if message is None:
+            raise FandangoValueError(
+                f"Party {self.party_name} received None-type message. This indicates a socket closing event. Overload the receive() method of {self.party_name} to handle this case."
```
riesentoaster:
```suggestion
                f"Party {self.party_name} received None-type message. This indicates a socket closing event. Override the receive() method of {self.party_name} to handle this case."
```

src/fandango/io/__init__.py
```diff
@@ -233,6 +237,7 @@ def __init__(
         self.port = port
         self.ip_type = ip_type
         self._party_instance = party_instance
+        self.io_instance = party_instance.io_instance
```
riesentoaster: This is a code smell. Why do we need this reference if I already have a reference to party_instance in these objects??

PR #824, opened 2026-02-25
feat: release 1.1.0rc2

joszamama:


riesentoaster:
Can you wait for #823, we're basically only waiting for the tests to finish there.

PR #842, opened 2026-02-27
Quickfix to make a CI test occupy less ram

alex9849:
Quick fix in order to make all CI tests running on Windows.
Set max_repetition_rate and max_nodes_rate to 0 in order to make the test occupy less memory.

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -28,4 +28,4 @@ jobs:
       # Run tests or coverage depending on branch and OS
       - name: Run tests
         run: |
-          pytest -v -n 2
+          pytest -v
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -28,4 +28,4 @@ jobs:
       # Run tests or coverage depending on branch and OS
       - name: Run tests
         run: |
-          pytest -v
+          pytest -v -n 2
--- a/tests/test_benchmarks.py
+++ b/tests/test_benchmarks.py
@@ -61,7 +61,9 @@ def test_generate_with_single_soft_constraint(benchmark: BenchmarkFixture):
         fan = Fandango._with_parsed(grammar, constraints)
         # make this a non-limiting factor
         max_generations = 10000
-        gen = fan.generate_solutions(max_generations=max_generations)
+        gen = fan.generate_solutions(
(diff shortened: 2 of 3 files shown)


.github/workflows/python-tests.yml
```diff
@@ -28,4 +28,4 @@ jobs:
       # Run tests or coverage depending on branch and OS
       - name: Run tests
         run: |
-          pytest -v
+          pytest -v -n 2
```
riesentoaster: `-n auto` is specified as the default in `pyproject.toml` — you're halving the execution speed with this.

tests/test_benchmarks.py
```diff
@@ -61,7 +61,9 @@ def func():
         fan = Fandango._with_parsed(grammar, constraints)
         # make this a non-limiting factor
         max_generations = 10000
-        gen = fan.generate_solutions(max_generations=max_generations)
+        gen = fan.generate_solutions(
+            max_generations=max_generations, max_repetition_rate=0.0, max_nodes_rate=0.0
+        )
```
riesentoaster, line 66: Why do we do this?

.github/workflows/python-tests.yml
riesentoaster: Regardless: If you think about changing this kind of thing, please talk to me. I have put significant thought into the testing and CI setup — there are reasons for these values.

alex9849: You can undo this part. I tried out a couple things because the windows ci runner was crashing. We want passing tests for the 1.1.0 release. But the fix has been to limit the max_nodes_rate and max_repetition_rate.

riesentoaster: @joszamama and I decided to just skip the test for now until we have a better understanding of what the problem actually is.

riesentoaster:
```suggestion
          pytest -v
```

riesentoaster:
I added an issue for the broken test in #845. If we find a solution for this, we can then close this issue. Do you believe the `max_repetition_rate` and `max_node_rate` changes fix the test? If yes: Any idea why?

.github/workflows/python-tests.yml
riesentoaster: (Sorry, I didn't see this soon enough.)

PR #848, opened 2026-03-02
Add case studies for PNG, GIF, and MP3

andreas-zeller:
This PR fixes a few minor bugs in the documentation, adds new case studies for PNG, GIF, and MP3; and adds a style guide.


docs/Style.md
```diff
@@ -0,0 +1,290 @@
...
+However, applying this style across the entire grammar can quickly lead to high indentation levels.
+If you break down your grammar into individual sections, have each section again begin on column 1.
+
+### Use Correct Fandango Syntax
+
+Here are the most important Fandango syntax rules:
+
+```
+     <NT> ::= ALT_1 | ALT_2    productions / alternatives
+     "..."   r'...'            string literal / regex
```
riesentoaster:
```suggestion
     "..."   r'...'  b"..."    string literal / regex / bytes object
```

docs/Style.md
```diff
@@ -0,0 +1,290 @@
...
+contexts `C`, give it a separate prefix for each context: `<C_checksum>` and `<C_length>`.
+
+
+### Cover the Entire Specification
+
+Do not just produce a toy subset, but go for maximum coverage of the respective language or protocol.
+Share only specifications that define the vast majority of format features.
+
+```{tip}
+Document missing parts with `TODO`, `FIXME` and alike.
```
riesentoaster:
```suggestion
Document missing parts with `TODO`, `FIXME` and alike and in the `Details` section at the beginning of the spec.
```

docs/Style.md
```diff
@@ -0,0 +1,290 @@
...
+### Follow Python Style Conventions
+
+Unless otherwise stated, stick to Python conventions for writing your specifications.
+This includes
+
+* code layout (4 spaces indentation; max line length 72 characters);
+* usage of blank lines (to separate larger grammar blocks);
+* source file encoding (should be UTF-8);
+* whitespace in expressions and statements; and
+* naming conventions for Python elements.
```
riesentoaster: We should at some point probably provide a linter for specs as well. Well, someone™ should.

docs/Style.md
```diff
@@ -0,0 +1,290 @@
...
+
+### Use Repetitions Wherever Possible
+
+Prefer `+` and `*` to denote repetitions, and `?` to denote optional elements.
+Avoid modeling repetition by means of the grammar.
+See {ref}`sec:repeat` for details.
+
+### Use Regular Expressions for Simple Elements
+
+For simple elements, prefer regular expressions (using `r"expr"`) over grammar forms.
```
riesentoaster: What denotes a `simple` element? An element that you don't care about exploring systematically?

docs/Style.md
```diff
@@ -0,0 +1,290 @@
...
+```
+
+where Fandango must make use of slower algorithms to find proper values for `<NT>` until `ok(...)` is finally satisfied.
+
+
+### Prefer Inline Expressions over Function Definitions
+
+Prefer a `where` clause or an inline Python expression directly in the grammar over introducing named Python functions.
+
+Function definitions are justified _only_ when the computation is genuinely too complex to write as one `where` expression or one `:=` expression inline (e.g. a real multi-step checksum algorithm) — never merely because a value is "random," reused in several places, or convenient to name.
```
riesentoaster, line 244: Also add a note about using multiple small `where` clauses over one big clause?

docs/Style.md
```diff
@@ -0,0 +1,290 @@
...
+```shell
+$ fandango fuzz -f SPEC.fan -n 10 -d out -x .EXTENSION --warnings-are-errors
+```
+
+For GIF files in `gif,fan`, this would be:
+```shell
+$ fandango fuzz -f gif.fan -n 10 -d out -x .gif --warnings-are-errors
+```
+
+Ensure the Fandango exit code is 0 _and_ `out/` has exactly 10 files — if not, fix and re-run.
```
riesentoaster:
```suggestion
Ensure the Fandango exit code is 0, Fandango does not print any warnings or error messages, _and_ `out/` has exactly 10 files — if not, fix and re-run.
```

riesentoaster (commented):
A couple of minor notes and questions

joszamama (commented):
I fixed the failing tests and pushed to this branch, and merged `main` in since it was 114 commits behind.

### The test failure

`gif89a.fan` includes `gif.fan` and refines symbols that came from the bt2fan-generated grammar (`<GifHeader>`, `<LogicalScreenDescriptor>`, `<UNDEFINEDDATA>`, `<ExtensionIntroducer_9>` and others). The new hand-written `gif.fan` defines none of them, so `TestGIFParsing::test_gif` failed on all twelve jobs with `Undefined symbol <LogicalScreenDescriptor> in grammar`.

I did not just rename the symbols, because the new `gif.fan` cannot parse GIF files at all, as `Gif.md` itself notes. Parsing `tinytrans.gif` with it gives `missing input at end of file`. So I kept the generated grammar as `gif-generated.fan` and pointed `gif89a.fan` at that. Your new `gif.fan` is untouched and stays the generation case study.

One thing worth knowing for the future: only `fandango parse` runs the grammar consistency check. `fandango convert` exits 0 on the broken file, which is why this surfaced in `test_parsing` rather than in the parser comparison test.

### wav.fan (removed)

I took it out. It was not referenced from any `.md` file or from `_toc.yml`, and it did not produce output in reasonable time: `fandango fuzz -f wav.fan -n 10 -d out -x .wav` gave 0 of 10 files in about 15 minutes, and even `-n 1` produced nothing within 2 minutes. For comparison, `gif.fan`, `png.fan` and `mp3.fan` each produce their 10 files in seconds and exit cleanly under `--warnings-are-errors`.

### Style.md

I like it a lot, and the principles are the right ones. I applied the small fixes directly:

* `gif,fan` is now `gif.fan`
* "checking helper func- tions" rejoined, it looked like a hyphenation artifact from a paste
* "max line length 72 characters" is now 88, matching our Ruff config. PEP 8 uses 79 for code and 72 for docstrings, so 72 contradicted both the standard and our own tooling
* the metadata example now says the capability lines describe what your spec actually does. As written, "The spec can parse any valid GIF file" read as a claim about `gif.fan`, which cannot parse GIFs
* `gif.fan` opened with `# gif_32x32.fan`, naming a file that does not exist and using an underscore where the guide asks for hyphens

Left for you, since these are your calls rather than mine:

* None of `gif.fan`, `png.fan` or `mp3.fan` carry the Authors / Sources / Details block the guide asks for. Since they ship alongside it, should they be the reference examples?
* The guide says to prefer inline expressions over function definitions, and specifically not to define one merely because a value is random. The new specs define 5, 4 and 1 helper functions respectively, several using `random`. Some, like the LZW encoder, are clearly covered by the "genuinely too complex" carve-out. Others like `int_to_le_bytes` feel closer to the line. Where would you put it?
* "Use `lowercase_identifiers` for nonterminals" conflicts with what we already ship. `Calculator.fan`, `fandango.fan`, `gif89a.fan`, `smtp-extended.fan` and `tinytrans.fan` all use CamelCase, and bt2fan emits CamelCase too. Should generated grammars be exempt, or do we want a migration at some point?
* The three sections on generators versus constraints ("Prefer Inline Expressions", "Prefer Constraints over Generators", "Use Generators only where Appropriate") make closely related points and could probably be one.

Nothing here blocks merging from my side.

joszamama (approved):

PR #874, opened 2026-03-27
Prevent runaway processes

riesentoaster:
Depends on #764

We shouldn't randomly start and then kill processes in the docs, but instead properly manage our processes.


riesentoaster:
(superseded by #894)

PR #889, opened 2026-04-15
Improve balancing across cores for unit tests

riesentoaster:
We want to run some tests first because they are very slow. Once they're done, we can fill the cores with the faster tests to not have to wait for the long ones at the end. This was already implemented. This PR improves how this is enforced and adds docs.


joszamama (approved):
LGTM

PR #893, opened 2026-04-20
Fix/generator updates read only

### LINKED ISSUE #892: Converter outputs are not marked read-only after parameter updates
When using a converter in a Fandango grammar like this:

```
<start> ::= <encrypted>
<encrypted> ::= <byte>+ := encrypt(<unencrypted>)
<unencrypted> ::= <structured_grammar> := decrypt(<encrypted>)
```

`<unencrypted>` represents a structured sub-grammar. During execution, `<unencrypted>` may be modified, for example during crossover. This triggers a recomputation of the `<encrypted>` generator value.

However, when the generator is re-run, the `DerivationTree` node for `<encrypted>` loses its read-only flag. As a result, later crossover mutations may incorrectly select one of `<encrypted>`’s children. Replacing such a child can produce semantically invalid input for the `decrypt` function, which then fails when deriving `<unencrypted>`.

alex9849:
Patch to fix Issue #892.

Closes #892

### DIFF
--- a/src/fandango/language/tree.py
+++ b/src/fandango/language/tree.py
@@ -860,6 +860,8 @@ class DerivationTree:
                 new_tree.sources = []
             else:
                 new_tree.set_children(grammar.derive_generator_output(new_tree))
+                for child in new_tree.children:
+                    child.set_all_read_only(True)
         elif regen_params:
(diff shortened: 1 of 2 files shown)


riesentoaster (approved):
Love that you added a unit test for it!

PR #899, opened 2026-04-22
Introduce `pylock.toml`

henryhchchc:
Fixes #873


henryhchchc:
`test_cli_min_2` appears to be a flaky test.

riesentoaster:
> `test_cli_min_2` appears to be a flaky test.

Yeah I've noticed this as well, but this only started being a problem recently. Probably worth investigating. That's another PR though. 

Generally: looks good. The only thing missing: should we add a note to the docs?

henryhchchc:
> should we add a note to the docs?

The docs already mentioned `pre-commit` and since we are extending our `pre-commit` hooks I think it should be fine?

riesentoaster:
> uv.lock is a snapshot of the (actual) dependencies used in this project. After updating the dependencies in pyproject.toml, make sure uv.lock is up-to-date with uv lock.

We have this note which suggests that adding the uv.lock is enough. I fear this may lead to confusion...

henryhchchc:
Yeah, I updated the docs as well.

docs/Contributing.md
```diff
@@ -144,8 +144,12 @@ This will run code formatting and type checking before every commit.
 The [tests](sec:running-tests) are not run as part of pre-commit, because they take significant time to complete. Run them manually.
 ```
 
-`uv.lock` is a snapshot of the (actual) dependencies used in this project.
-After updating the dependencies in `pyproject.toml`, make sure `uv.lock` is up-to-date with `uv lock`.
+To support reproducible environments and compatibility across a wider range of tools in the Python ecosystem, we keep both `uv.lock` and `pylock.toml` up to date.
+
+- `uv.lock` is the lockfile generated by `uv` and captures the actual dependency set used in this project (i.e., source of truth).
```
riesentoaster:
```suggestion
- `uv.lock` is the lockfile generated by `uv` and captures the actual dependency set used in this project (i.e., source of truth). Update with `uv lock`
```

docs/Contributing.md
```diff
@@ -144,8 +144,12 @@ This will run code formatting and type checking before every commit.
...
+- `pylock.toml` is derived from `uv.lock` to support a wider range of ecosystem tools.
```
riesentoaster:
```suggestion
- `pylock.toml` is derived from `uv.lock` to support a wider range of ecosystem tools. Update with `uv export --all-extras --locked --output-file=pylock.toml`.
```

docs/Contributing.md
```diff
@@ -144,8 +144,12 @@ This will run code formatting and type checking before every commit.
...
+
+After updating dependencies in `pyproject.toml`, make sure to regenerate `uv.lock` with `uv lock` and update `pylock.toml` accordingly.
```
riesentoaster:
```suggestion
After updating dependencies in `pyproject.toml`, make sure to regenerate `uv.lock` and `pylock.toml` to keep them up to date. Both pre-commit hooks and the CI pipeline contain checks for these.
```

riesentoaster (commented):
I would consider leaving both update commands explicitly here

riesentoaster (approved):
Looks good now! Thank you for the work on this and the other administrative parts — Fandango is becoming a pretty mature project from that point of view, which is great to see.

PR #902, opened 2026-04-28
remove exception catch in io method

alex9849:
This PR removes an exception catch from the generate_io method. This error should be given to the user.


riesentoaster (approved):
Man, we really need to refactor and split this function 😬

alex9849:
Yes, I agree. I'm already on it. I just wanted this to be merged first.

PR #915, opened 2026-05-12
Remove Dependabot `pylock.toml` update workflow

henryhchchc:
As per discussion with @riesentoaster, we decided to disable the auto update of `pylock.toml` for the moment (due to [CI issues](https://github.com/fandango-fuzzer/fandango/pull/901#issuecomment-4368757980)). The dependabot update interval is already 7 days BTW.


riesentoaster:
> The dependabot update interval is already 7 days BTW.

I know — my comment via DM was more about the fact that we're getting individual PRs for all non-patch updates. Which means that in some weeks, we may need to manually fix the `pylock.toml` in like 5 PRs. That could get annoying. We could also make dependabot only create one PR with all updates together. This is much more likely to fail, but if we'll have to do some manual work anyways, I don't think this would be that bad.

That's just a suggestion though.

henryhchchc:
Yeah right – I grouped the minor updates as well. For major updates, I think we should still test/review them one by one.

.github/dependabot.yml
```diff
@@ -8,6 +8,9 @@ updates:
     schedule:
       interval: weekly
     groups:
+      minor-updates:
+        update-types:
+          - "minor"
       patch-updates:
         update-types:
           - "patch"
```
riesentoaster, line 16:
```suggestion
      minor-updates:
        update-types:
          - "minor"
          - "patch"
```

riesentoaster: Should we merge these?

riesentoaster:
> For major updates, I think we should still test/review them one by one.

Sure, makes sense.

.github/dependabot.yml
henryhchchc: This will make a singe PR for all the minor and patch updates. Is this what we want?

riesentoaster: I don't know, which is why I'm asking. I would probably only keep them separate if we expect these to break at least sometimes. Otherwise splitting them just forces us to do some work twice. But I have no data to support this question.

henryhchchc: Minor and patch updates have different likelihood for breaking things. So I think having two update groups is a good balance between the burden and the risk – at least we reduced the amount of work from one PR per dependency.

henryhchchc:
Ok the flaky tests are green!

.github/dependabot.yml
riesentoaster: Fair enough.

riesentoaster (approved):

PR #928, opened 2026-05-27
Introduce Ruff for both formatting and linting

riesentoaster:
I got annoyed by unused imports and variables, and wrong orders in imports. So obviously the only reasonable thing to do (*sigh*) is add more tooling.

Also: This will raise errors if one uses mutable structures as default parameters in functions, as these are only created once and re-used in future calls. This has already lead to problems at some point.


Makefile
```diff
@@ -325,6 +325,10 @@ ls-files:
 	@echo 'Listing files in the repository...'
 	@$(GIT_LS_FILES) | sort
 
+lock:
+	uv lock
+	uv export --output-file=pylock.toml --all-extras --locked
+
```
riesentoaster, line 331: @henryhchchc you may be interested in this.

henryhchchc: Nice! This is convenient.

henryhchchc:
That's a nice direction towards cleaner code! We may also consider [Ruff](https://docs.astral.sh/ruff/). It serves as both a linter and a formatter, which covers the functionalities of both flake8 and black, and it's quite mature. Also it runs much more faster. It's build by the authors of UV BTW.

riesentoaster:
> That's a nice direction towards cleaner code! We may also consider the [Ruff](https://docs.astral.sh/ruff/). It serves as both a linter and a formatter, which covers the functionalities of both flake8 and black, and it's quite mature. Also it runs much more faster. It's build by the authors of UV BTW.

I would love that, but moving linters can be a bit of a pain because everyone needs to update their systems. But if you feel like doing it, you have my full support :)

henryhchchc:
> moving linters can be a bit of a pain because everyone needs to update their systems

Yes that was the reason why I hesitated. Now this PR gives a good chance -- we can use ruff for the roles of both flake8 and balck. Would you mind if I push directly to this PR?

riesentoaster:
Please do!

I have also "only" added two flake8 plugins, and there are many more possibly worth adding. But this PR is already quite large, so I wanted to stick to the minimum. Certainly worth considering though.

In the medium term, I would love some sort of automated code review that enforces some best practices that go beyond simple regex checkers. But there's a very fine balance between enforcing too many things and then people just start ignoring it/want to remove it. Hmm…

riesentoaster:
Btw: The main downside of astral projects is that [they joined OpenAI](https://astral.sh/blog/openai). Wo while for now it seems like a good solution, I am not quite sure how much trust I want to put into the longevity of its stability due to presumably an imminent and large amount of vibe coded changes. I guess we can always pin versions though.

henryhchchc:
The changes are landed and we got speedup -- the ruff checks finish nearly instantly. Let's see how the tests go.

> The main downside of astral projects is that [they joined OpenAI](https://astral.sh/blog/openai).

Ouch, the AI companies are nowadays trying to acquire everything (bun was also acquired by Anthropic). I think it would be difficult to entirely stay away from this if the AI bubble keeps growing 🙃.

riesentoaster:
> The changes are landed and we got speedup -- the ruff checks finish nearly instantly. Let's see how the tests go.

Nice! Now we should just™ find a way from always writing integration tests and move to more scoped unit tests to speed them up. Not sure that's going to go anywhere though. 

> > The main downside of astral projects is that [they joined OpenAI](https://astral.sh/blog/openai).
> 
> Ouch, the AI companies are nowadays trying to acquire everything (bun was also acquired by Anthropic). I think it would be difficult to entirely stay away from this if the AI bubble keeps growing 🙃.

I guess this is now just an unfortunate reality of the world we live in.

henryhchchc:
> find a way from always writing integration tests and move to more scoped unit tests to speed them up

Yes, also to reduce flakiness – the flaky tests are restarted.

riesentoaster:
> > find a way from always writing integration tests and move to more scoped unit tests to speed them up
> 
> 
> 
> Yes, also to reduce flakiness – the flaky tests are restarted.

💯

Having some integration tests makes sense for sure. But most things shouldn't be tested by running the main algorithm. 

Oh, and while we're at it, adding a bunch of extra benchmarks for smaller sections would be very nice. Things like tree evaluation, parsing, etc would be obvious first targets.

henryhchchc:
Nice now the tests are green.

> Oh, and while we're at it, adding a bunch of extra benchmarks for smaller sections would be very nice. Things like tree evaluation, parsing, etc would be obvious first targets.

Totally agree. We should get three things done at some point. (a) refactor the algorithm into several self-contain, testable components, (b) have unit tests for these components while integration tests become lightweight smoke tests, and (c) benchmarks for hot-paths. I think that would take several PRs.

riesentoaster:
Agreed.

This PR is ready to merge from my side, please check my work and review/merge.

henryhchchc (approved):
LGTM! Let's get this landed.

PR #934, opened 2026-06-08
Make cache size specifiable with an environment variable

riesentoaster:


alex9849 (changes requested):
PR looks good for me. But some tests are failing

riesentoaster:
Not sure how I managed to trigger copilot, but that was a mistake. Force pushed back to my changes.

@alex9849 the CI is unfortunately quite unstable ever since the last merging spree. I'm not sure what the problem is. I'll just keep rerunning them until they pass.

edit: but I'll do that from your branch to not have to do that twice.

PR #936, opened 2026-06-09
Introduce --format=1

riesentoaster:
This is useful for testing when you're just interested in how many inputs get produced when by Fandango.


Namron2000 (approved):
Looks good

PR #937, opened 2026-06-09
Improve Dockerfile

riesentoaster:
- Adapt to build system changes over the last months (even year?)
- Adopt multi-stage build which reduces image size by ~90%


joszamama (approved):
lgtm

PR #949, opened 2026-07-01
Introduce Permutations and Timers

alex9849:


src/fandango/evolution/algorithm/algorithm_selector.py
```diff
@@ -0,0 +1,23 @@
+from fandango.evolution.algorithm.base import GeneticAlgorithm
+from fandango.language import Grammar
+from fandango.language.grammar import FuzzingMode
+
+
+class AlgorithmSelector:
```
riesentoaster: Why is this done here? Why don't we select this wherever `FuzzingMode` is set and just call the appropriate constructor there?

Can we just get rid of `FuzzingMode` all together? This requires annoying checks in other places they don't belong (e.g. in `SimpleAlgorithm`).

src/fandango/evolution/algorithm/base.py
```diff
@@ -68,3 +69,7 @@ def generate(
         mode: FuzzingMode = FuzzingMode.COMPLETE,
     ) -> Generator[DerivationTree, None, None]:
         pass
+
+    @abstractmethod
+    def reset(self) -> None:
+        pass
```
riesentoaster, line 75: Document what this is supposed to do.

src/fandango/api.py
```diff
@@ -263,9 +270,22 @@ def init_population(
...
+                self.fandango = ProtocolAlgorithm(
+                    packet_algorithm=SimpleGeneticAlgorithm(
+                        grammar=self.grammar,
+                        constraints=constraints,
+                        start_symbol=start_symbol,
+                        **settings,
+                    ),
+                )
+            case _:
+                raise ValueError(f"Unknown fuzzing mode: {self._last_fuzzing_mode}")
```
riesentoaster, line 290: r/e the comment above: Here we're already switching outside the algorithm classes. This is what should also be done in the CLI.

src/fandango/io/navigation/graph/visitor/continuing_nodevisitor.py
```diff
@@ -138,7 +141,8 @@ def visitAlternative(self, node: Alternative) -> bool:
                 try:
                     continue_exploring = self.visit(alt)
                     found = True
-                    break
+                    # if not node.is_permutation:
+                    #    break
```
riesentoaster: What's up with this?

src/fandango/io/__init__.py
riesentoaster, line 1: (FYI: I didn't look at IO internals, those are your own responsibility :D)

src/fandango/language/grammar/nodes/alternative.py
```diff
@@ -19,10 +19,12 @@ def __init__(
         alternatives: list[Node],
         grammar_settings: Sequence[HasSettings],
         id: str = "",
+        is_permutation: bool = False,
     ):
         assert len(alternatives) > 0, "alternatives must be non-empty"
         self.id = id
         self.alternatives = alternatives
+        self.is_permutation = is_permutation
```
riesentoaster: Do I see it correctly that this is the only part of the internal language we needed to change? That would be _very nice_.

riesentoaster: One nitpick: Is this read-only? Then we should mark it as such by keeping the value private and writing only a getter. (Yes, that is not enforced by Python itself, but it is the pythonic way of marking it.)

src/fandango/language/parse/convert.py
```diff
@@ -269,7 +271,37 @@ def visitRepeat(self, ctx: FandangoParser.RepeatContext):
...
+            branches.append(self._concat_two_nodes(first, tail))
+        self.seenAlternatives += 1
+        nid = self.seenAlternatives
+        return Alternative(
+            branches,
+            self._grammar_settings,
+            f"{NodeType.ALTERNATIVE}:{nid}_{self.id_prefix}",
+            is_permutation=True,
+        )
+
```
riesentoaster, line 305: Can you document what happens here?

src/fandango/language/parse/parse_tree.py
```diff
@@ -46,7 +47,23 @@ def parse_tree(filename: str, fan_contents: str) -> ParseTree:
...
+                )
+                input_stream = InputStream(fan_contents)
+                saved = sa_fandango.USE_CPP_IMPLEMENTATION
+                sa_fandango.USE_CPP_IMPLEMENTATION = False
+                try:
+                    tree = sa_fandango.parse(input_stream, "fandango", error_listener)
+                finally:
+                    sa_fandango.USE_CPP_IMPLEMENTATION = saved
+            else:
+                raise
```
riesentoaster: Do we need to put this somewhere in the docs/the CLI help messages/_somewhere_?

riesentoaster (changes requested):
I'm impressed! Looks pretty good!

Very minor comments only.

Oh, and please rename your PRs to something more descriptive. And add a short description of what it does if it isn't immediately obvious.

src/fandango/language/grammar/nodes/alternative.py
alex9849: Yep, that's basically the only change internally.

src/fandango/language/parse/parse_tree.py
alex9849: This shouldnt be here at all. This comes from me mergine main into my branch. It looks like something got overwritten during the merge

alex9849:
@riesentoaster Thanks for the review. I addressed you requested changes.

src/fandango/io/navigation/graph/visitor/continuing_nodevisitor.py
alex9849: Shouldn't be there. That's an artifact. Commenting this out is a bugfix. I deleted the code in my revision.

src/fandango/evolution/algorithm/algorithm_selector.py
alex9849: The class is referenced nowhere. I deleted it. Seems to be dead code anyway.

src/fandango/api.py
alex9849: This is the CLI. Isn't it?

riesentoaster: Nope, that's the API. There we probably shouldn't change the interface, so this is fine.

src/fandango/evolution/algorithm/base.py
riesentoaster: What's up with this?

riesentoaster: Ah, sorry, I didn't have all commits selected in the diff 🤦‍♂️

src/fandango/io/navigation/graph/stategrammarconverter.py
```diff
@@ -86,7 +86,7 @@ def visitAlternative(self, node: Alternative) -> Alternative:
             self.visitChildren(node),
             self._grammar_settings,
             node.id,
-            is_permutation=node.is_permutation,
+            is_permutation=node._is_permutation,
```
riesentoaster: Don't access a private element from the outside. Instead, access as normal:

```suggestion
            is_permutation=node.is_permutation,
```

src/fandango/language/grammar/nodes/alternative.py
```diff
@@ -24,7 +24,7 @@ def __init__(
         assert len(alternatives) > 0, "alternatives must be non-empty"
         self.id = id
         self.alternatives = alternatives
-        self.is_permutation = is_permutation
+        self._is_permutation = is_permutation
         super().__init__(NodeType.ALTERNATIVE, grammar_settings)
```
riesentoaster, line 29: And then here, do this:
```suggestion
        self._is_permutation = is_permutation
        super().__init__(NodeType.ALTERNATIVE, grammar_settings)

        @property
        def is_permutation(self):
                return self._is_permutation
```

If you don't add the corresponding setter, IDEs will complain if you try to write to this field.

src/fandango/io/navigation/selection/protocol_model.py
```diff
@@ -66,7 +66,7 @@ def _build_permutation_groups(
     def _collect_permutation_groups(
         self, node: Node, groups: dict[NonTerminal, frozenset[NonTerminal]]
     ) -> None:
-        if isinstance(node, Alternative) and node.is_permutation:
+        if isinstance(node, Alternative) and node._is_permutation:
```
riesentoaster: And then here also

```suggestion
        if isinstance(node, Alternative) and node.is_permutation:
```

riesentoaster (changes requested):

riesentoaster:
One final nitpick

alex9849:
@riesentoaster Thankyou. I integrated that. I also merged the changes from main back into my branch. 👍

riesentoaster (approved):
Very nice! Thanks for the work! Love that you added tests as well!

riesentoaster:
(I have changed the PR title for something a bit more expressive. Feel free to adjust it again if you want to frame this differently, but these titles end up in e.g. the release notes, so they should probably be somewhat representative.)

PR #952, opened 2026-07-07
Move to unified build system

riesentoaster:
#950 and #951 break. So I spent some time to clean things up. No more manual `setup.py`!

One dirty hack necessary to keep beartype working (upstream issue exists already).


henryhchchc (approved):
Cool! LGTM. It's much cleaner!

PR #953, opened 2026-07-09
Remove additional import finder

riesentoaster:
As suggested here: https://github.com/beartype/beartype/issues/556#issuecomment-4918427368

### DIFF
--- a/src/fandango/beartype.py
+++ b/src/fandango/beartype.py
@@ -1,6 +1,9 @@
 def _is_fandango_editable_redirect_finder(finder: object) -> bool:
     """True for this package's scikit-build-core redirect editable finder only."""
-    if type(finder).__name__ != "ScikitBuildRedirectingFinder":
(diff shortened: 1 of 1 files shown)


henryhchchc (approved):

henryiii:
Did this one actually cause any problems? It delegates to PathFinder and should handle this just fine? (Also, are you even using it? What king of configuration do you have that uses both inplace mode and redirect mode? Normally you can't have both, but I guess with overrides, or if you manually override?)

riesentoaster:
> Did this one actually cause any problems? It delegates to PathFinder and should handle this just fine? (Also, are you even using it? What king of configuration do you have that uses both inplace mode and redirect mode? Normally you can't have both, but I guess with overrides, or if you manually override?)

I don't *think* we're using it, this was more a "for completeness' sake" triggered by the comment in the beartype issue thing. Worked before this change, continued working after.

PR #955, opened 2026-07-09
Move to maturin as a Build System

riesentoaster:
Reason: `scikit-build-core` was just creating too many problems. We still want a parallel build. As a nice side effect, this also prepares the ability to write hot-path performance-critical parts in Rust.

Changes:
- Build the parser with cargo's `cc` module
- Write a small Rust wrapper around the C++ parser to make it accessible to Python
- Introduce a dummy Rust function to show access for future functions
- Ensure type synchronization between Rust and Python
- Remove the no longer needed `FANDANGO_SKIP_CPP_PARSER` (and `FANDANGO_REQUIRE_BINARY_BUILD`) — we always build on developer machines and will have to do this as soon as we move anything serious to Rust; consumers download binaries anyway, they don't need the toolchain


riesentoaster:
(CodeQL is still broken for now because it can no longer auto-detect the build system. We would need to either just drop it or provide a custom config, but that's definitely doable.)

PR #957, opened 2026-07-10
Use maturin + CMake as build system

riesentoaster:
Just checking if this is better.


riesentoaster:
I don't think this is simpler; worth a try but oh well. If anything, we should go the full maturin (#955) way.

PR #975, opened 2026-08-03
Fix --warnings-are-errors

riesentoaster:
- `--warnings-are-errors` previously didn't raise exceptions in e.g. constraints but just print them as warnings.
- We were also needlessly passing around a `warnings_are_errors` argument through the algorithm and nested classes. Cleaned up here.

### DIFF
--- a/src/fandango/api.py
+++ b/src/fandango/api.py
@@ -13,7 +13,7 @@ from fandango.language.grammar import FuzzingMode, ParsingMode
 from fandango.language.grammar.grammar import Grammar
 from fandango.language.parse.parse import parse
 from fandango.language.tree import DerivationTree
-from fandango.logger import LOGGER
+from fandango.logger import LOGGER, set_raise_on_logged_exceptions
 
 DEFAULT_MAX_GENERATIONS = 500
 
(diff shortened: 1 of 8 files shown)


alex9849 (commented):
You removed that `self._warnings_are_errors` attribute from many classes, but only in the constructor. Are they never used in the class's code for actual logging?

riesentoaster:
> You removed that `self._warnings_are_errors` attribute from many classes, but only in the constructor. Are they never used in the class's code for actual logging?

Not as far as I can tell. But you may know something I don't (maybe about IO?).

alex9849:
> > You removed that `self._warnings_are_errors` attribute from many classes, but only in the constructor. Are they never used in the class's code for actual logging?
> 
> Not as far as I can tell. But you may know something I don't (maybe about IO?).

I guess, if the whole thing runs, it's not referenced. This may have been a dead attribute all along.

Edit: I just checked the code. It's really not referenced. I'll approve this.

riesentoaster:
> This may have been a dead attribute all along.

That's what I figured as well.

alex9849 (approved):

PR #978, opened 2026-08-03
Rework the contributing guide and add the community health files

joszamama:
The contributing guide had drifted from the repository, and most of the files GitHub expects a project to have were missing.

## What was broken

Three instructions in `docs/Contributing.md` did not work:

- it told you to install a **`evaluation` extra that does not exist** (the real ones are `book`, `development`, `full`, `test`, and `full` was never mentioned),
- the Windows activation path was `. venv/Scripts/activate`, missing a dot,
- a stray `ti` sat on its own line in the middle of the install steps.

Structurally, the reliable path was buried. Steps 3 to 5 walked you through `venv` plus `pip install -e .`, while CI uses `uv sync --all-extras --locked` and we maintain `uv.lock` as the source of truth. `uv` appeared only in a tip beginning "If subsequent tests fail due to dependency breakage". The guide also never stated which Python versions we support.

## The guide, rewritten

- **uv first**, matching CI exactly, with pip kept as a genuine alternative and the C++ parser fallback (`FANDANGO_SKIP_CPP_PARSER=1`) documented where you hit it
- **a table of what CI checks.** Formatting and types run over the whole repository, not just `src`, which is the most common reason a pull request goes red
- **how to report a bug**, how the lockfiles are regenerated, and what makes a pull request easy to review
- **issues are claimed before work starts.** If an issue is assigned, leave it. If it is not, ask in the issue and wait to be assigned. The old text said the opposite
- **a section on AI assistance.** No objection to the tools. The requirement is that you own what you submit and can explain it in review, because a patch nobody has read costs more to review than the issue costs to fix
- contributions are licensed under **EUPL-1.2**, same as the project

## New files

| File | Why |
|---|---|
| `CONTRIBUTING.md` | root copy, which is what GitHub actually surfaces on a new PR or issue |
| `CODE_OF_CONDUCT.md` | the guide cited a code of conduct with no file behind it and no way to report a violation |
| `SUPPORT.md` | routes usage questions away from the bug tracker |
| `CITATION.cff` | GitHub can now offer the ISSTA 2025 paper, validated against the CFF 1.2.0 schema |
| `.editorconfig` | matches the Ruff settings, tabs for `Makefile` |
| `.github/PULL_REQUEST_TEMPLATE.md` | |
| `.github/ISSUE_TEMPLATE/` | bug report, feature request, and a config routing security reports away from public issues |

## Security policy

`SECURITY.md` now routes reports through GitHub's **private advisory form** rather than a shared mailbox, so the report, our questions, the fix, and the advisory stay in one private thread, and we can request a CVE and credit the reporter from it. The mailbox survives as a fallback, with a note not to put exploit details in a first email.

It also gains response times and a **scope** section. Fandango executes the specifications it is given, by design, so running an untrusted spec is not a vulnerability. Saying that up front is easier than arguing it case by case.

## README

Gains an install command, which it did not have at all, plus links to the guide, the code of conduct, support, the security policy, the hands-on tutorial, and how to cite.

## Checks

Every command and extra in the new guide was verified against the Makefile and `pyproject.toml`. All documentation links resolve, `CITATION.cff` validates against the CFF schema, the issue templates parse, and the book builds.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,8 +1,9 @@
-# EditorConfig helps keep formatting consistent across editors.
+# EditorConfig keeps basic file hygiene consistent across editors.
 # See https://editorconfig.org
 #
-# These settings match the Ruff configuration in pyproject.toml, so your editor
-# and `ruff format` agree. Ruff remains the source of truth for Python.
+# Deliberately minimal. Ruff is the source of truth for Python formatting
+# (indentation, line length), so those are not repeated here: duplicating them
+# only creates two places to disagree. This file covers what Ruff does not.
 
 root = true
 
@@ -12,25 +13,14 @@ end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 indent_style = space
-indent_size = 4
-
-[*.py]
-indent_size = 4
-max_line_length = 88
-
-[*.{yml,yaml,json,toml}]
-indent_size = 2
 
 [*.{md,fan}]
-indent_size = 4
-# Markdown uses trailing double-space for hard line breaks, and .fan specs are
-# whitespace-significant, so leave trailing whitespace alone in both.
+# Markdown uses a trailing double space for hard line breaks, and .fan specs
+# are whitespace-significant.
 trim_trailing_whitespace = false
 
-[*.g4]
-indent_size = 4
-
 [Makefile]
+# Tabs are required here, not a preference.
 indent_style = tab
 
 [*.{bat,ps1}]
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,37 @@
+# EditorConfig helps keep formatting consistent across editors.
+# See https://editorconfig.org
+#
+# These settings match the Ruff configuration in pyproject.toml, so your editor
+# and `ruff format` agree. Ruff remains the source of truth for Python.
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+indent_size = 4
+
+[*.py]
+indent_size = 4
+max_line_length = 88
+
+[*.{yml,yaml,json,toml}]
+indent_size = 2
+
+[*.{md,fan}]
+indent_size = 4
+# Markdown uses trailing double-space for hard line breaks, and .fan specs are
+# whitespace-significant, so leave trailing whitespace alone in both.
+trim_trailing_whitespace = false
+
+[*.g4]
+indent_size = 4
+
+[Makefile]
+indent_style = tab
+
+[*.{bat,ps1}]
+end_of_line = crlf
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -46,7 +46,9 @@ body:
     id: output
     attributes:
       label: The full output
-      description: The complete error message and traceback, not just the last line.
+      description: >-
+        The complete error message and traceback, not just the last line.
+        Running the command with `-v` adds useful debugging information.
       render: shell
     validations:
       required: true
@@ -56,7 +58,7 @@ body:
     attributes:
       label: Fandango version
       description: The output of `fandango --version`.
-      placeholder: Fandango 1.1.1
+      placeholder: Fandango x.x.x
     validations:
       required: true
 
@@ -64,7 +66,7 @@ body:
     id: python-os
     attributes:
       label: Python version and operating system
-      placeholder: Python 3.13.5 on macOS 15.2
+      placeholder: Python 3.x.x on macOS x.x
     validations:
       required: true
 
@@ -75,5 +77,5 @@ body:
       options:
         - label: I searched the existing issues and did not find this
           required: false
-        - label: I am using a supported Python version (3.11 or later)
+        - label: I checked that this still happens on the current `main` branch
           required: false
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,79 @@
+name: Bug report
+description: Something in Fandango does not behave the way it should
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to report this.
+
+        **Found a security vulnerability?** Please do not file it here.
+        Report it privately with GitHub's [security advisory form](https://github.com/fandango-fuzzer/fandango/security/advisories/new).
+        See [SECURITY.md](https://github.com/fandango-fuzzer/fandango/blob/main/SECURITY.md).
+
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What happened?
+      description: What did you expect, and what did you get instead?
+      placeholder: |
+        I expected Fandango to generate inputs matching my grammar, but it
+        exited with an error instead.
+    validations:
+      required: true
+
+  - type: textarea
+    id: spec
+    attributes:
+      label: The spec
+      description: >
+        The smallest `.fan` spec that still shows the problem. Reducing it is
+        the single most useful thing you can do to get this fixed quickly.
+      render: text
+    validations:
+      required: false
+
+  - type: textarea
+    id: command
+    attributes:
+      label: The command you ran
+      placeholder: fandango fuzz -f my_spec.fan -n 10
+      render: shell
+    validations:
+      required: true
+
+  - type: textarea
+    id: output
+    attributes:
+      label: The full output
+      description: The complete error message and traceback, not just the last line.
+      render: shell
+    validations:
+      required: true
+
+  - type: input
+    id: version
+    attributes:
+      label: Fandango version
+      description: The output of `fandango --version`.
+      placeholder: Fandango 1.1.1
+    validations:
+      required: true
+
+  - type: input
+    id: python-os
+    attributes:
+      label: Python version and operating system
+      placeholder: Python 3.13.5 on macOS 15.2
+    validations:
+      required: true
+
+  - type: checkboxes
+    id: checks
+    attributes:
+      label: Before submitting
+      options:
+        - label: I searched the existing issues and did not find this
+          required: false
+        - label: I am using a supported Python version (3.11 or later)
+          required: false
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -1,0 +1,51 @@
+name: Feature request
+description: Suggest something Fandango should be able to do
+labels: ["enhancement"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for the suggestion. Describing the problem you ran into is more
+        useful than describing a solution, because it leaves room for an
+        approach you may not have considered.
+
+  - type: textarea
+    id: problem
+    attributes:
+      label: What problem are you trying to solve?
+      description: >
+        What were you doing when you needed this? What did you try, and where
+        did it fall short?
+      placeholder: >
+        I am specifying a format where a field's value depends on a checksum of
+        two other fields, and I could not find a way to express that.
+    validations:
+      required: true
+
+  - type: textarea
+    id: proposal
+    attributes:
+      label: What would you like Fandango to do?
+      description: If you have a concrete idea, sketch it. A `.fan` snippet helps.
+    validations:
+      required: false
+
+  - type: textarea
+    id: alternatives
+    attributes:
+      label: Alternatives you considered
+      description: >
+        Including "I worked around it like this", which is useful for judging how
+        much the feature would buy.
+    validations:
+      required: false
+
+  - type: checkboxes
+    id: checks
+    attributes:
+      label: Before submitting
+      options:
+        - label: I searched the existing issues and did not find this
+          required: false
+        - label: I would be willing to work on this myself
+          required: false
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -5,14 +5,14 @@ body:
   - type: markdown
     attributes:
       value: |
-        Thanks for the suggestion. Describing the problem you ran into is more
-        useful than describing a solution, because it leaves room for an
+        Thanks for the suggestion. Telling us what you were trying to do is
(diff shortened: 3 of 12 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,79 @@
...
+      placeholder: fandango fuzz -f my_spec.fan -n 10
+      render: shell
+    validations:
+      required: true
+
+  - type: textarea
+    id: output
+    attributes:
+      label: The full output
+      description: The complete error message and traceback, not just the last line.
```
riesentoaster: \+ ```Consider running your command with `-v`for additional debugging information.```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,79 @@
...
+      required: true
+
+  - type: input
+    id: version
+    attributes:
+      label: Fandango version
+      description: The output of `fandango --version`.
+      placeholder: Fandango 1.1.1
+    validations:
+      required: true
```
riesentoaster, line 63: Instead just add a checkbox with the other that people should validate that the bug exists in the current GitHub main branch HEAD version.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,79 @@
...
+
+  - type: input
+    id: python-os
+    attributes:
+      label: Python version and operating system
+      placeholder: Python 3.13.5 on macOS 15.2
```
riesentoaster: Make the placeholder something that obviously needs to change? e.g. `Python 3.x.x on macOS x.x`?

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,51 @@
+name: Feature request
+description: Suggest something Fandango should be able to do
+labels: ["enhancement"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for the suggestion. Describing the problem you ran into is more
+        useful than describing a solution, because it leaves room for an
+        approach you may not have considered.
```
riesentoaster: I wouldn't use the word problem here, problems should go into bug reports.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Checklist
+
+- [ ] The change does one thing
+- [ ] Tests cover any changed behaviour
+- [ ] Documentation updated, if this changes something a user can see
+- [ ] `pre-commit` passes (`pre-commit run --all-files`)
+- [ ] `make tests` passes locally (note the plural)
+- [ ] `make lock` was run, if `pyproject.toml` dependencies changed
+- [ ] I have read every line of this change and can explain it in review
```
riesentoaster: Are we going to expect these things to hold for ourselves as well? :D

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,53 @@
...
+already assigned to someone else, we will most likely close this in favour of
+their work, which is a waste of your time.
+-->
+
+
+<!--
+CI checks formatting and types across the WHOLE repository, not just src/, so a
+stray file elsewhere can turn the build red. Running pre-commit locally catches
+almost all of it.
+-->
```
riesentoaster: This is redundant.

.editorconfig
riesentoaster, line 1: Is there any way we can get away without specifying all these values here as well? This seems like a recipe for future inconsistencies.

CODE_OF_CONDUCT.md
riesentoaster, line 1: Love it, also love that we're not inventing our own.

CONTRIBUTING.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Reporting a bug
+
+Open an [issue](https://github.com/fandango-fuzzer/fandango/issues/new/choose)
+and include:
+
+- the smallest `.fan` spec that still shows the problem, and the exact command
+  you ran,
+- what you expected, and what happened instead, with the full error output,
+- the output of `fandango --version`, plus your Python version and OS.
```
riesentoaster, line 27: One thing I toyed around with literally today: Could we make `--version` be more specific if it isn't the released version? So it would e.g. report the commit on a checkout etc.? I think the build system supports something like `dynamic = ["version"]` (needs a bit more config, not sure if this is what we'd actually want), but could be useful.

CONTRIBUTING.md
```diff
@@ -0,0 +1,114 @@
...
+
+If you later cannot finish it, say so and unassign yourself. That is fine.
+
+## Using AI assistance
+
+We are not against AI tools, we use them too, and we do not ask you to declare
+them. But **you own the code you submit**: you should be able to explain why it
+is written that way, why it is correct, and what you rejected. If a reviewer
+asks about a decision and the honest answer is "the model wrote it", the pull
+request is not ready.
```
riesentoaster: 1. People should declare AI usage. I'd much rather know where the code came from, what kind of stability/insight I can expect etc. We cannot enforce it, I know, but it makes triage of PRs easier I think. See e.g. https://github.com/beartype/beartype/issues/556#issuecomment-4931334006 and later comments in that issue.
2. Would it make sense to ask people to write the PR description themselves/have that not LLM generated? LLMs tend to be very verbose which is a lot of work. And writing the description yourself forces you to actually understand what is happening. Again: We can't prove this, but as a guideline I think this would be worth doing.

CONTRIBUTING.md
```diff
@@ -0,0 +1,114 @@
...
+
+The problem is not AI, it is moving your work onto the maintainers. A patch
+generated from an issue and pushed unread costs us more time than the issue
+would have. Those pull requests get inspected closely, and if the work has been
+handed to us rather than done, we close them.
+
+Read every line, run it, and be ready to defend it in review. Do that and we do
+not care how you got there.
```
riesentoaster: I wouldn't say defend, that's too antagonistic to volunteers. Maybe something along the lines of
```
Make sure you understand what every changed line does and why it is there.
```

And if we say that we don't care, that's a hard invite for slop. I wouldn't mention that. It's not necessary, we say the same thing above already.

CONTRIBUTING.md
```diff
@@ -0,0 +1,114 @@
...
+make tests           # note the plural: `make test` does nothing
+```
+
+CI checks formatting and types across the **whole repository**, not just `src`,
+so a stray file elsewhere can fail the build. `pre-commit` catches almost all
+of it locally.
+
+If you change dependencies in `pyproject.toml`, regenerate both lockfiles with
+`make lock`, or CI will fail.
+
```
riesentoaster, line 110: This should link to the docs instead in case we change this later. Otherwise we might miss one of the locations and break things again.

CITATION.cff
riesentoaster, line 1: Should we also add a bibtex note somewhere?

SECURITY.md
```diff
@@ -1,33 +1,86 @@
...
+   which you prefer.
+3. Release a fixed version and note the advisory in the
+   [release notes](https://fandango-fuzzer.github.io/ReleaseNotes.html).
+
+## Scope
+
+Fandango is a **testing tool**. It runs specifications you provide, and a `.fan`
+specification can contain arbitrary Python by design, which it executes. Running
+an untrusted specification is therefore equivalent to running untrusted code,
+and that is expected behaviour rather than a vulnerability.
```
riesentoaster, line 66: Good idea lol, Fandango is basically RCE as a service.

.github/PULL_REQUEST_TEMPLATE.md
riesentoaster, line 1: I would add a note that the diff should be minimal. LLMs have a tendency to touch *everything*, and that's really annoying to review.

riesentoaster (commented):
Good stuff! I left a couple of comments, some suggestions, some questions. Most of them are fairly loosely held. Three overall things:

- The templates are quite verbose. We should possibly trim those down to something short. Keep in mind: Every time we open a PR or an issue, we'll have to go through it as well.
- I would like to emphasis my comment regarding the LLM usage policy: I would ask for declaration (even though we can't enforce it), and I would ask people to write their issues/PR descriptions manually.

.github/PULL_REQUEST_TEMPLATE.md
joszamama: So for this specifically, I have been reporting some bugs in open source projects and they have this kind of thing. And it is quite good ngl I like it. It also makes clear many things that you should do in case you are running some agents

SECURITY.md
joszamama: Yeah lol I can see "hey I ran fandango with this grammar and now my system does not respond, here you have my grammar: `<start> ::= execute('sudo rm -rf /')`. Help!"

.github/ISSUE_TEMPLATE/bug_report.yml
joszamama: Added, as part of the description on that field: "Running the command with `-v` adds useful debugging information."

.github/ISSUE_TEMPLATE/bug_report.yml
joszamama: Done. Replaced the Python version checkbox with "I checked that this still happens on the current `main` branch". Better signal, and it saves us reproducing something already fixed.

.github/ISSUE_TEMPLATE/bug_report.yml
joszamama: Done, now `Python 3.x.x on macOS x.x`. Did the same for the Fandango version field, which said `Fandango 1.1.1` and would have gone stale anyway.

.github/ISSUE_TEMPLATE/feature_request.yml
joszamama: Fair, changed. The field is now "What are you trying to do?" and the intro asks what you wanted Fandango to do rather than what problem you hit. Renamed the field id too so nothing says "problem".

.github/PULL_REQUEST_TEMPLATE.md
joszamama: Removed.

.github/PULL_REQUEST_TEMPLATE.md
joszamama: Added. The checklist item is now "Does one thing, and the diff is limited to what that needs", and the AI section in the guide says the same with your reasoning: models reformat and rename things the change never needed to touch, and every unrelated hunk is more for a reviewer to rule out.

.editorconfig
joszamama: Good catch, trimmed. It now only covers what Ruff does not: encoding, line endings, final newline, tabs for the Makefile, and leaving trailing whitespace alone in .md and .fan since both are whitespace-significant. Indentation and line length are gone, so Ruff stays the single source of truth. I left a comment in the file saying exactly that, so nobody re-adds them.

CONTRIBUTING.md
joszamama: Both in. The section now asks people to say when a change was substantially AI-assisted, framed so it does not read as an accusation: we cannot verify it, it does not count against you, and it tells a reviewer what kind of read the change needs. It also asks people to write the description themselves, with your reasoning that generated ones are long, say little, and skip the part where you work out whether you actually understand the change.

CONTRIBUTING.md
joszamama: Agreed on both, and I used your wording. "Be ready to defend it in review" is now "Make sure you understand what every changed line does and why it is there". I also cut "we do not care how you got there"; you are right that it reads as an invitation for slop, and it repeated the point made two paragraphs earlier.

CONTRIBUTING.md
joszamama: Done. The root file now links to the lockfile section of the guide instead of restating the commands. Same reasoning applies to the rest of that file, which is deliberately a short pointer to the book rather than a second copy of it.

CITATION.cff
joszamama: Added the full entry to the README, next to the "Cite this repository" note. I checked it field by field against docs/fandango.bib so the two cannot drift, and kept the note that the first author family name is "Zamudio Amaya", since tools tend to mangle it.

CONTRIBUTING.md
joszamama: Not doing this one here, but I like it. Making `--version` report the commit for a checkout is a change to the build and to the tool, not to the policy, so it does not belong in a docs-and-templates PR. It also needs a real decision (`dynamic = ["version"]` via setuptools-scm or similar, and what we want it to print for an editable install off a dirty tree), which deserves its own thread rather than being settled in passing here.

It would help the bug template, though: right now we ask for `fandango --version` and get `Fandango 1.1.1` from someone running a checkout that is 200 commits ahead. Worth an issue, and I am happy for you to open it since it was your idea.

andreas-zeller (approved):

.github/PULL_REQUEST_TEMPLATE.md
riesentoaster: Don't get me wrong: I love it! It's just, …, some of our PRs tend to be _huge_.

CONTRIBUTING.md
riesentoaster: Yeah, for sure, it's a different thing. Just wanted to see if you'd be open in principle. I'll open an issue then.

Edit: #981

.github/PULL_REQUEST_TEMPLATE.md
joszamama: Ik, but that is something we should be trying not to do in the future. I think this could help, and if it doesn't, we can always roll it back.

joszamama: Something that could happen is that we do not adhere to it very well because we push whole features at a time, but for external contributors (and notably, reviewing their PRs) this will def. help

CONTRIBUTING.md
```diff
@@ -0,0 +1,126 @@
...
+never needed to touch, and every unrelated hunk is more for a reviewer to rule
+out.
+
+## Setting up
+
+Fandango needs **Python 3.11 or later**. With
+[uv](https://docs.astral.sh/uv/) (recommended, this is what our CI uses):
+
+```shell
+git clone https://github.com/YOUR-USERNAME/fandango.git
```
riesentoaster, line 80: Should we add a note about the process?
1. Fork the Fandango repository
2. Clone your fork
3. Change what you want to, commit and push to your fork
4. Create a PR across forks to the main Fandango repository

PR #985, opened 2026-08-10
Grammars with unbreakable loops now cause an error instead of letting Fandango being stuck in an infinite recursion

### LINKED ISSUE #926: Priming Grammar is broken with recursive grammars
Grammars such as

```
<start> ::= <start>
```

run into an infinite loop with no error message. We should detect this and raise an exception.

(Initial report by Norman, based on constructs such as `<deep_arr> ::= '[' <deep_arr> ']' := '[' + ']'`)

alex9849:
## What and why

Fixes https://github.com/fandango-fuzzer/fandango/issues/926

Instead of using a loop for priming, we now have a visitor that does the distance_to_completion computation.
(Easier to understand in my opinion). We break the stuck-in-priming issue by introducing a "changed" flag that, for each "priming" iteration, notes whether that iteration changed any distance_to_completion value. If nothing changed, priming is finished. Afterwards, we can filter out all nodes with infinite distance_to_completion. These then represent unbreakable loops, which let fandango not terminate before this commit. In case we detect such a loop, we terminate with an error and point to the non-terminals part of that unbreakable loop.

## How was it tested?

new unittest

## Checklist

      substantially AI-assisted

### DIFF
--- a/src/fandango/language/grammar/node_visitors/primer.py
+++ b/src/fandango/language/grammar/node_visitors/primer.py
@@ -85,7 +85,7 @@ class PrimerVisitor(NodeVisitor[float, float]):
         return self.visitRepetition(node)
 
     def visitNonTerminalNode(self, node: NonTerminalNode) -> float:
-        return self.visit(self._rules[node.symbol]) + 1
+        return 1 + self.visit(self._rules[node.symbol])
 
     def visitTerminalNode(self, node: TerminalNode) -> float:
         return 1.0
--- a/src/fandango/language/grammar/node_visitors/primer.py
+++ b/src/fandango/language/grammar/node_visitors/primer.py
@@ -1,0 +1,76 @@
+from fandango.language.grammar.node_visitors.node_visitor import (
+    NodeVisitor,
+    ResultType,
+)
+from fandango.language.grammar.nodes.alternative import Alternative
+from fandango.language.grammar.nodes.char_set import CharSet
+from fandango.language.grammar.nodes.concatenation import Concatenation
+from fandango.language.grammar.nodes.node import Node
+from fandango.language.grammar.nodes.non_terminal import NonTerminalNode
+from fandango.language.grammar.nodes.repetition import Option, Plus, Repetition, Star
+from fandango.language.grammar.nodes.terminal import TerminalNode
+from fandango.language.symbols.non_terminal import NonTerminal
+from fandango.language.symbols.symbol import Symbol
+
+
+class PrimerVisitor(NodeVisitor):
+
+    def __init__(self, rules: dict[NonTerminal, Node]):
+        self._rules = rules
+        self._seen_count: dict[Node, int] = dict()
+        self.inf_loops: set[Node] = set()
+
+    def prime(self):
+        self._seen_count.clear()
+        self.inf_loops.clear()
+        for rule in self._rules.values():
+            self.visit(rule)
+        self.inf_loops = {n for n in self.inf_loops if n.distance_to_completion == float('inf')}
+
+    def visit(self, node: Node) -> ResultType:
+        seen_count = self._seen_count.get(node, 0)
+        if seen_count > 1:
+            return float('inf')
+        self._seen_count[node] = seen_count + 1
+        dist = super().visit(node)
+        node.distance_to_completion = dist
+        if dist == float('inf'):
+            self.inf_loops.add(node)
+        self._seen_count[node] = self._seen_count.get(node, 0) - 1
+        return dist
+
+    def visitAlternative(self, node: Alternative) -> ResultType:
+        minimal_dist = float('inf')
+        for child in node.children():
+            child_dist = self.visit(child)
+            if minimal_dist > child_dist:
+                minimal_dist = child_dist
+        return minimal_dist + 1
+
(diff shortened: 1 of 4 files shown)


src/fandango/language/grammar/node_visitors/primer.py
```diff
@@ -0,0 +1,82 @@
...
+        self._seen.remove(id(node))
+        return dist
+
+    def visitAlternative(self, node: Alternative) -> float:
+        minimal_dist = float("inf")
+        for child in node.children():
+            child_dist = self.visit(child)
+            if minimal_dist > child_dist:
+                minimal_dist = child_dist
+        return minimal_dist + 1
```
riesentoaster:
```suggestion
        return 1 + min(self.visit(child) for child in node.children())
```

src/fandango/language/grammar/node_visitors/primer.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def visitConcatenation(self, node: Concatenation) -> float:
+        total_dist = 0.0
+        for child in node.children():
+            total_dist += self.visit(child)
+        return total_dist + 1
```
riesentoaster:
```suggestion
        return 1 + sum(self.visit(child) for child in node.children))
```

src/fandango/language/grammar/node_visitors/primer.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def visitRepetition(self, node: Repetition) -> float:
+        child_dist = self.visit(node.node)
+        if node.min == 0:
+            return 1.0
+        return (node.min * child_dist) + 1
```
riesentoaster: No need for the manual check, the product turns to 0 if node.min is 0.

```suggestion
        return 1 + (node.min * self.visit(node.node))
```

src/fandango/language/grammar/node_visitors/primer.py
```diff
@@ -0,0 +1,82 @@
...
+    def visit(self, node: Node) -> float:
+        if id(node) in self._seen:
+            return node.distance_to_completion
+        self._seen.add(id(node))
+        dist = super().visit(node)
+        if node.distance_to_completion > dist:
+            self._changed = True
+            node.distance_to_completion = dist
+        if dist == float("inf"):
+            self.inf_loops.add(node)
```
riesentoaster: We aren't sure this is inf yet, right? So we add it and then potentially remove it again later? Why not only do detection in the end?

src/fandango/language/grammar/node_visitors/primer.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class PrimerVisitor(NodeVisitor[float, float]):
+    def __init__(self, rules: dict[NonTerminal, Node]):
+        self._rules = rules
+        self._seen: set[int] = set()
+        self.inf_loops: set[Node] = set()
+        self._changed = False
+
+    def prime(self) -> None:
```
riesentoaster: Add docs

src/fandango/language/grammar/node_visitors/primer.py
```diff
@@ -0,0 +1,82 @@
...
+from fandango.language.grammar.nodes.char_set import CharSet
+from fandango.language.grammar.nodes.concatenation import Concatenation
+from fandango.language.grammar.nodes.node import Node
+from fandango.language.grammar.nodes.non_terminal import NonTerminalNode
+from fandango.language.grammar.nodes.repetition import Option, Plus, Repetition, Star
+from fandango.language.grammar.nodes.terminal import TerminalNode
+from fandango.language.symbols.non_terminal import NonTerminal
+
+
+class PrimerVisitor(NodeVisitor[float, float]):
```
riesentoaster, line 13: Add docs

src/fandango/language/grammar/node_visitors/primer.py
```diff
@@ -0,0 +1,82 @@
...
+        self._changed = False
+        for rule in self._rules.values():
+            self.visit(rule)
+        while self._changed:
+            self._changed = False
+            for rule in self._rules.values():
+                self.visit(rule)
+        self.inf_loops = {
+            n for n in self.inf_loops if n.distance_to_completion == float("inf")
+        }
```
riesentoaster: I'm still in favor of just raising an exception here. If you are really concerned that somebody might not want to abort because of this, you can always make it a custom exception, document that it might happen, and a caller can catch it.

riesentoaster (changes requested):
Love the added tests.

src/fandango/language/grammar/node_visitors/primer.py
```diff
@@ -0,0 +1,94 @@
...
+        return self.visitRepetition(node)
+
+    def visitPlus(self, node: Plus) -> float:
+        return self.visitRepetition(node)
+
+    def visitOption(self, node: Option) -> float:
+        return self.visitRepetition(node)
+
+    def visitNonTerminalNode(self, node: NonTerminalNode) -> float:
+        return self.visit(self._rules[node.symbol]) + 1
```
riesentoaster:
```suggestion
        return 1 + self.visit(self._rules[node.symbol])
```

riesentoaster (approved):

PR #989, opened 2026-08-12
Fix: unreachable-symbol error when a packet symbol expands to another packet symbol

alex9849:
## What and why

Fixed the bug described in issue #860:

Consider the following grammar:

```
<start> ::= <Party:msg_a><Party:msg_b>
<msg_a> ::= 'example'
<msg_b> ::= <msg_a>
```

This grammar can lead to a navigation error with the error message:
```
raise FandangoError(fandango.errors.FandangoError: Symbol (NonTerminal(<start>), NonTerminal(<_packet_msg_b>), NonTerminal(<_packet_msg_a>)) is not reachable in grammar.
```

This happens because `<msg_b>` expands to `<msg_a>`, while both `<msg_a>` and `<msg_b>` appear in contexts annotated with a sending party. As a result, the navigation logic may treat both as packet-level symbols. When computing a route for the corresponding k-path, the system may attempt to reach a sequence involving both packetized symbols even though such a path is not actually reachable in the grammar.

## How was it tested?

New unittest

## Checklist

      substantially AI-assisted

### DIFF
--- a/src/fandango/io/navigation/selection/protocol_model.py
+++ b/src/fandango/io/navigation/selection/protocol_model.py
@@ -18,6 +18,7 @@ class ProtocolModel:
     def __init__(self, grammar: Grammar, start_symbol: NonTerminal):
         self._grammar = grammar
         self.state_grammar_symbols = self._get_state_grammar_symbols(start_symbol)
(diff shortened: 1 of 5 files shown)


riesentoaster (approved):
Didn't look through the changes in the IO-specific code, that's on you. Rest looks good.

Issue #370, opened 2025-03-15
Parsing time increases substantially for longer strings


andreas-zeller:
Try the following (in `tests/long_string.fan`):

```python
<start> ::= <char>+ := "a" * 100
```

and then

```shell
$ time fandango fuzz -f tests/resources/long_string.fan -n 1 --population-size=1
```

which should result in the constructor being called once, and then the parser called once.

I note that as the length of the string increases, parsing length increases much more than linearly (quadratically? exponentially?):

100 - 0.64s
200 - 1.03s
300 - 1.87s
400 - 3.62s
500 - 6.41s
600 - 10.62s
700 - 16.19s
800 - 23.77s
900 - 33.48s
1000 - 45.06s

Please check on this - the Earley parser should not show such behavior.

alex9849:
I have done some experiments and tried to get behind this problem.
It looks like the problem occurs when trying to parse long repetitions. This is not a bug. This is the worst case scenario when using an early parser that has a worst case runtime complexity of O(n^3).
The "good" news is that this only happens within a single repetition. When a repetition ends, the algorithm starts to speed up again. For example. This is a simplified grammar similar to the example you provided:
```
<start> ::= "a"+ := "a" * 300
```
For me this runs in 1,37s.


This is a similar example. Now with a nested repetition, that contains the repetition from above, but with a "b", that "brakes" the inner repetition and starts a new one.
```
<start> ::= ("a"+ "b")+ := (("a" * 300) + "b") * 10
```
On my machine this example runs in 10.06s, although there are 10 times more letters to parse.

A repetition that contains a static number of repetitions. Runs even faster.
```
<start> ::= "a"{3000} := "a" * 3000
```
This runs in 1,8s on my machine. Even though its repetition parses 3000 letters.


I looked at the state of the parsing table and how it evolves with increasing repetition values.
It looks like repetitions only run with such a bad runtime complexity when the amount of how often the repetition should be repeated is uncertain. For a grammar with `<start> ::= "a"{2500, 3000} := "a" * 3000`, the repetitions are uncertain for `3000 - 2500 = 500` times, because it doesn't know in advance how many "a"s will come. And it is only for these 500 that the complexity becomes so bad.


[...]

PR #997, opened 2026-08-31
A faster parser

alex9849:
## What and why

The parser is still somewhat slow. This PR introduces some opmimizations to speed it up.
Introduced optimizations:

### 1. Leo shortcuts
When completing the element of a parser rule the parser just cycles over it, adds the now complete state to the current column, and repeats that for every item waiting on the symbol. Usually that's a handful. With right recursion it isn't. Every input position leaves another item waiting, so one completion cascades up the entire stack to the current column ends up holding one completed item per origin. This produces O(n^2) items for certain rules. (e.g. +, *, and {N, M}-rules).

With the Leo optimizations, we can add a shortcut for the last symbol of a parsing state that points directly to the parsing state that started the chain of rules. Complete therefore now runs in O(1) instead of O(n^2) for those rules. We had something like that before already, but this addition here optimizes these shortcuts further and introduces them dynamically to all rules where they can apply.

### 2. Edges instead of parsed children
A `ParseState` used to carry the derivation trees built so far. That made `complete()` copy the whole child list. So for each complete() step, the entire so-far parsed children list got copied. (Running in O(n)).
A ParseState now holds an edge that points from the new ParseState to the old one. It holds only the parsed symbol/DerivationTree that got parsed in the step between the new and the old parse state. So we only store the parsing delta. 

At the end of our parsing session, can traverse this "graph" of parse states through all combinations and construct the final derivation tree in O(n) (per DerivationTree instance). 
In case of ambiguous parsing, instead of getting a combinatorial explosion in parsing states, we can just add a single second edge between one and another ParseState. When constructing the derivation tree, each different combinatorial option to traverse our ParseState-graph is a new DerivationTree. So each new combination of one DerivationTree might only be one Edge between ParseStates, instead of a copy of the entire parsing process. The state graph grows only polynomially with an exponentially growing number of parsed trees. This saves us a lot of memory and time.
Constructing parsingTrees from this graph form is also possible iteratively. Yielding one or many DerivationTree doesn't require more parsing overhead, but just requires the algorithm to construct more derivationTrees from the ParseState-graph. 

### 3. Disabled garbage collection during parsing.
Python's cyclic garbage collector frequently scans every live container object to look for reference cycles, and the parse table is full of objects. Each time it runs, it scans our entire parsing table, and the parsing table is huge.
So instead of letting it run over the parse table continuously, we disable the collector for the duration of a parse and enable it again afterwards, where it cleans up the old table as usual.

## How was it tested?

Comparing to the numbers from 2025 made in 2025: #370 

It used this grammar to test how fast Fandango's parser is:
```
<start> ::= <char>+ := "a" * 100
```

We compare the initial parser used before that referenced issue, with the parser with the optimizations from that issue, with the parser that uses the optimizations from this PR.

String length | Initial parser | Old optimized parser | This PR (looking like O(n) for this case)
-- | -- | -- | --
100 | 0.64s | 0.13s | 0.003s
200 | 1.03s | X | 0.006s
300 | 1.87s | 0.57s | 0.009s
400 | 3.62s | X | 0.012s
500 | 6.41s | 0.85s | 0.015s
600 | 10.62s | X | 0.017s
700 | 16.19s | 1.41s | 0.020s
800 | 23.77s | X | 0.023s
900 | 33.48s | X | 0.026s
1_000 | 45.06s | 3.13s | 0.03s
1_500 | X | 10.01s | 0.044s
2_000 | X | 25s | 0.057s
10_000 | X | X | 0.31s
50_000 | X | X | 1.71s
100_000 | X | X | 3.47s
200_000 | X | X | 7.17s
500_000 | X | X | 18.93s
1_000_000 | X | X | 41.82s

## Checklist

      substantially AI-assisted


riesentoaster:
Would it make sense to do some differential testing here between the old and new parser; and in general for pre/post parsing?

For this, I'd take all specs in the repo (check the list for the spec parser unit test), have Fandango produce a single tree from each, serialize and parse, and check if it is still the same (or at least, because some specs might have multiple valid parses for each linearized one, any parsed tree is the same as the original one, and all of them are correct w.r.t. the grammar/constraints). This would for now, and for the future, not just test for an absolutely trivial spec (as you used in your benchmark) but every weird edge case we put in any other unit test. 

Also: could you use some spec and create a benchmark for it in the benchmark unit tests? That would protect us from performance regressions in the future.

riesentoaster:
Love the optimizations btw! Cool stuff!

src/fandango/evolution/crossover.py
```diff
@@ -23,7 +23,8 @@ def crossover(
         common_symbols = symbols1.intersection(symbols2)
         if not common_symbols:
             return None
-        symbol = random.choice(list(common_symbols))
+        # Sorted so the choice is independent of PYTHONHASHSEED.
+        symbol = random.choice(sorted(common_symbols, key=lambda s: s.name()))
```
riesentoaster: We shouldn't sort here; this is an unnecessary performance hit. If someone needs a fixed seed, they should manually set PYTHONHASHSEED, as is documented with the `--random` command line option.

alex9849: Are you sure? This fixed a unit test with soft constraints for me that was unreliable.

andreas-zeller:
First, let me chime in on the praise for these optimizations. Good work!

The `--validate` flag in the Fandango CLI (activated for most of the tutorial examples) ensures that whatever is produced parses into the same derivation tree again. This has helped a lot in achieving parser quality. Generally speaking, if this invariant is never violated across tens of thousands of produce/parse pairs, one can very much rely on the parser doing the right thing™.

If you want to do differential testing, you could also extend this validation to compare the old and new parsers (assuming there are ways to keep both active); then you can achieve near-100% certainty in their correctness.

alex9849:
I already ran most of our grammars with the --validate flag from the CLI. That worked, but I didn't do direct differential testing with the old parser, and I also added a few more changes yesterday after that `--validate` test. I'll see if I can find a way to install an old and a new Fandango version on my system for that. Shouldn't be too difficult. :)

Edit: I actually found a bug during incomplete parsing. Working on it.

src/fandango/evolution/crossover.py
riesentoaster: Is it possible that the var isn't properly propagated to subcommands? I took great pain at some point to fix this for the unit tests.

alex9849:
So, I added a few more things.
The branch now includes parser benchmarking. 
I also ran differential testing sessions against the old parser. The parser behaves in the same way, and I even fixed a couple of bugs that the old parser had. I did not commit these differential tests.

During the benchmarking process, however, I noticed that the parser can quickly encounter a recursion error when working with highly nested grammars. I therefore refactored the parser to process states iteratively instead of recursively. Getting this right took a while. This also includes a couple of methods in grammar.py and tree.py that now run iteratively.

The parser's foundation is also much more powerful now. 
Our usual parser in parser.py has always been a wrapper for iterative_parser.py. This provided a simplified interface to a much more powerful parser, which has become even more powerful.

iterative_parser.py was better in that it could continue an already started parse with a word that should be appended to the previously given word, without restarting the parsing process. This is very useful for protocol testing. However, it can now also detect potential complete parses in a word that is longer than a complete parse. So a word that starts with a byte string that contains but extends a complete parse. This means that we can detect message borders between multiple incoming network messages without consuming all incoming packets byte-wise. This speeds up packet parsing by a factor of O(n).

The .consume() method no longer produces trees. The method is no longer a generator. It only processes inputs now. From there, a user can ask the parser for certain properties of the parse, such as:
- "Can I continue this parse?"
- "Is this parse complete?"
- "Where are potential message borders?"
Now these questions can all be answered without producing the trees. This saves a lot of overhead, as users are not required to construct trees to answer these questions themselves.

I also added a couple of unit tests.

@riesentoaster could you take a look again? I guess the parsing logic is a bit too complicated at this point, but everything around that and the interface should be reviewable.

src/fandango/language/symbols/non_terminal.py
```diff
@@ -5,15 +5,17 @@ class NonTerminal(Symbol):
     def __init__(self, symbol: str) -> None:
         assert isinstance(symbol, str)
         super().__init__(symbol, SymbolType.NON_TERMINAL)
+        self._name = symbol
```
riesentoaster: Aren't we storing the same data in two places (here and in the super class)? That seems inefficient and unnecessary. We should either have them always in the children (if there's no such thing as a generic `symbol` with a consistent semantic) or in the parent (if there is).

src/fandango/language/symbols/non_terminal.py
```diff
@@ -5,15 +5,17 @@ class NonTerminal(Symbol):
...
+        self._hash = hash((self._value, self._type))
```
riesentoaster: Is this actually a necessary optimization? This breaks if any object is mutated and adds complexity.

src/fandango/language/tree.py
```diff
@@ -255,15 +258,12 @@ def get_path(self) -> list["DerivationTree"]:
...
-        for child in self._children:
-            child.set_all_read_only(read_only)
-        for child in self._sources:
-            child.set_all_read_only(read_only)
+        stack = [self]
+        while stack:
+            node = stack.pop()
+            node.read_only = read_only
+            stack.extend(node._children)
+            stack.extend(node._sources)
```
riesentoaster, line 270: Where did the docs go? Also check `invalidate_hash` please.

src/fandango/language/tree.py
```diff
@@ -420,16 +420,28 @@ def __hash__(self) -> int:
...
+        stack: list[DerivationTree] = [self]
+        while stack:
+            node = stack.pop()
+            if node.hash_cache is None:
+                order.append(node)
+                stack.extend(node._children)
+
+        result = 0
+        for node in reversed(order):
+            result = node.hash_cache = hash(
```
riesentoaster: This isn't very clean. Couldn't we just return `self.hash_cache` in the end without having to constantly reassign this?

src/fandango/language/tree.py
```diff
@@ -420,16 +420,28 @@ def __hash__(self) -> int:
         """
         Computes a hash of the derivation tree based on its structure and symbols.
         """
-        if self.hash_cache is None:
-            self.hash_cache = hash(
+        if self.hash_cache is not None:
+            return self.hash_cache
+
+        order: list[DerivationTree] = []
+        stack: list[DerivationTree] = [self]
```
riesentoaster: This is not a very expressive variable name, changing this to something like `to_recompute` and `non_cached` (or anything else you come up with!) might make this code more understandable.

tests/test_benchmarks.py
```diff
@@ -1,16 +1,189 @@
...
+<key> ::= r'[a-z]+'
+<value> ::= <number> | <quoted> | <empty>
+<number> ::= <digit>{1,4}
+<digit> ::= '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9'
+<quoted> ::= '"' <char>* '"'
+<char> ::= 'x' | 'y'
+<empty> ::= ''
+<flags> ::= <flag>?
+<flag> ::= '!'
+"""
```
riesentoaster: Could we move these to actual files? They seem useful for other tests (at the very least the spec parser tests).

tests/test_benchmarks.py
riesentoaster, line 1: There seems to be _a lot_ of complexity for a benchmark. Is this actually necessary? Could we simplify?

src/fandango/evolution/algorithm/simple.py
```diff
@@ -424,7 +424,9 @@ def _generate_simple(
                 new_population = self._perform_destruction(new_population)
 
             # Ensure Uniqueness & Fill Population
-            new_population = list(set(new_population))
+            # dict.fromkeys keeps the population order independent of
+            # PYTHONHASHSEED, so a fixed --random-seed stays reproducible.
+            new_population = list(dict.fromkeys(new_population))
```
riesentoaster: This shouldn't be necessary. If the `PYTHONHASHSEED` isn't fixed, other things will subtly break as well, we should fix the root cause, not do this here.

src/fandango/evolution/crossover.py
riesentoaster: Ditto here, except this notably adds a bunch of unnecessary compute.

riesentoaster (changes requested):
I didn't take a look at the parser internals nor the IO things — those are your territory and it would take me too long to understand it to be reasonable. I can only ask that you do the work properly and make sure it's correct with all tools we have at our disposal. The differential tests and the added unit tests seem to be a good start though — thank you for that!

tests/test_benchmarks.py
alex9849: It's not that complex. If you look at the code, it really just creates the parser and doesn't perform really long parses for normal unit tests, but only during benchmarking. But I can move the grammars to the .fan files. I thought they would only be used there.

src/fandango/language/symbols/non_terminal.py
alex9849: It's not a quadratic optimization, but without it the parser is slowed down by about 10%. We can get rid of this, but I thought since all this ir read-only it would be no big deal.

src/fandango/language/symbols/non_terminal.py
alex9849: I did this change together with the hashing one. But yes, thre the benefix is negectible. I changed it back. (Pushing after resolving the other comments.)

src/fandango/language/tree.py
alex9849: This method computes the hash value for the entire subtree, without calling the hash function of the other children. I rewrote this to prevent recursion errors in deep trees.

We first go down the derivation tree depth-first, and add them to a list. This is now the flat ordered list where the parents of all children in a tree come first.

We can then traverse that list from the back and compute the hash value of those parents/children. By doing this, we always get a cache hit when computing the parent, since the children are processed first. This prevents us from getting a recursion error for deep trees.

But when thinking about this, maybe I should use the flatten method for that and use that one here.

tests/test_benchmarks.py
alex9849: yes.

src/fandango/language/tree.py
alex9849: I replaced it with a different method that I now refactored to also work iteratively and use that here.

src/fandango/language/tree.py
riesentoaster: That's not what I meant – but I'm very much open for improvements that simplify this further. This was a statement specifically about this line:

```python
            result = node.hash_cache = hash(
```

alex9849:
Maybe one point. I disabled beartype during benchmarking. That made the benchmarking time increase by a factor of 4, and we already run beatype during the normal pytest tests.

riesentoaster:
> Maybe one point. I disabled beartype during benchmarking. That made the benchmarking time increase by a factor of 4, and we already run beatype during the normal pytest tests.

Disabling beartype for benchmarking seems reasonable. And I assume that _decreased_ the time, not increased it, right? Also: I'm pretty sure all benchmarks are also run as part of the ordinary unit tests, but only once without timing etc.

src/fandango/language/symbols/non_terminal.py
riesentoaster: For now it is, but it isn't documented or enforced. This is just a silent expectation and those are recipes for future failures. We may want to look into something like `@dataclass(frozen=True)`, possibly also with `slots=True` for even more speedups.

riesentoaster: (Also put those in the super class if you do go that route, or at least check. I'm not super familiar with dataclasses)

src/fandango/language/tree.py
```diff
@@ -420,16 +424,22 @@ def __hash__(self) -> int:
         """
         Computes a hash of the derivation tree based on its structure and symbols.
         """
-        if self.hash_cache is None:
-            self.hash_cache = hash(
+        if self.hash_cache is not None:
+            return self.hash_cache
+
+        flatted = self.flatten()
```
riesentoaster: (nitpick: should be `flattened`)

tests/test_benchmarks.py
```diff
@@ -1,16 +1,162 @@
...
+    word = case.make_word(length)
+
+    def func():
+        _run_case(case, word, fandango, grammar)
+
+    try:
+        benchmark.pedantic(
+            func,
+            setup=grammar._parser._cache.clear,
+            iterations=1,
```
riesentoaster, line 155: We're manually forcing exactly one repetition for this benchmark, where for the others, the repetitions are automatically calculated, and provide more accurate results for fast benchmarks.

Is there any sort of randomness in parsing? If yes, we need multiple repetitions.

src/fandango/language/tree.py
alex9849: done

src/fandango/language/symbols/non_terminal.py
alex9849: okay, moved to code to symbol.

Namron2000 (approved):
Looks good to me, great job!

